PR #110, opened 2014-11-18
Update basic.md

fasar:
Abort can be received by a Router to abort a session before an established session.


oberstet:
@beatgammit yep, the thinking was as you outline: client send `HELLO`, router sends `CHALLENGE` and client chokes by sending `ABORT`

Paranaix:
It should be noted that a client might also decide to abort if the server simply can't respond in time or refuses to do so (e.g caused by buggy code) but doesn't drop the transport.

Issue #145, opened 2015-03-16
extra field in WELCOME message


Paranaix:
For some usecases it might be desireable to transmit arbitrary data upon successful authentication. Such data might for example be a session token and its lifetime. 

This issue proposes to add an optional *extra* field to the *WELCOME* message, which is a dictionary. The content of yonder dictionary is completely up to the user/router and can be anything.  Empty dictionaries should be allowed. If none *extra* field is present, WAMP client implementations should treat this case as like an empty dictionary was transmitted.

A modified version of the *WELCOME* message in [Example 3](]https://github.com/tavendo/WAMP/blob/master/spec/advanced.md#example-3) which incorporates the proposed changes might look like this:

```javascript
[
   2, 
   7286787554810878, 
   {
      "authid": "tobias.oberstein@gmail.com", 
      "authmethod": "mozilla_persona", 
      "authrole": "com.tavendo.community.role.user", 
      "roles": {
         "broker": {
            "features": {
               "publisher_exclusion": true, 
               "publisher_identification": true, 
               "subscriber_blackwhite_listing": true
            }
         }, 
         "dealer": {
            "features": {
               "caller_identification": true, 
               "progressive_call_results": true
            }
         }
      }
     "extra": {
       "token": "936DA01F-9ABD-4D9D-80C7-02AF85C822A8",
       "until": 123456789,
       "someOtherAttribute": 42
     }
   }
]
```

oberstet:
Ah, ok, makes sense.

This is pretty straighforward also to add to Autobahn:

1. Extending the WELCOME message https://github.com/tavendo/AutobahnPython/blob/master/autobahn/wamp/message.py#L341
2. Extending the SessionDetails https://github.com/tavendo/AutobahnPython/blob/master/autobahn/wamp/types.py#L200 that get provided to https://github.com/tavendo/AutobahnPython/blob/master/autobahn/wamp/interfaces.py#L284

Similar changes are required for AutobahnJS and AutobahnCpp. In routers, support will also be required. E.g. in Crossbar.io, the static and dynamic authentication facilities need to be enriched for this extra dictionary and exposed at user level.

PR #147, opened 2015-03-23
Implemented #145

Paranaix:
Solves #145

### DIFF
--- a/spec/basic.md
+++ b/spec/basic.md
@@ -462,7 +462,7 @@ Sent by a *Client* to initiate opening of a WAMP session to a *Router* attaching
 
 Sent by a *Router* to accept a *Client*. The WAMP session is now open.
 
-    [WELCOME, Session|id, Details|dict]
+    [WELCOME, Session|id, Details|dict, Extra|dict]
 
 #### ABORT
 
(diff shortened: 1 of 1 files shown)


Paranaix:
Any comments on this one? Should this be implemented in crossbar first as proof of concept before merging?

oberstet:
@Paranaix This doesn't follow your original proposal of just adding an `extra` attribute in the already existing `WELCOME.details|dict`, but extends the message itself. There is no need for the latter.

Paranaix:
Maybe you missunderstood me then in the first place. Practically whether its seperate or in `details`doesn't make any difference. However I prefer an additional field instead of one monster of a dictionary. But to be honest, I dont really care. If this is a problem, I can change it as well.

However notice that the standard should explicitely state which fields are in the `details` (and any other dictionaries as well). At the moment many aspects of the standard are quite "interpreteable" if people only consult the standard and not the reference implementations (autobahn and crossbar).

Paranaix:
Why did you closed this? Especially without a reply? Sorry but sometimes I don't get the way you work. I told you why I prefer it the current way, but at the same time tell you that thats not very important for me and I could change it if you still insist on doing it another way. I wait and wait for a reply, but instead of giving me one you suddenly close this PR...

ecorm:
I'm not taking sides here, but I'd like to point out that adding an `extra` field to the `WELCOME` message might break existing implementations that bother to check the schema of incoming WAMP messages. I know for sure that it would break my C++ library (CppWAMP). The whole idea behind the `Options` and `Details` fields was to be able to extend WAMP without breaking implementations.

oberstet:
@Paranaix 

> Practically whether its seperate or in detailsdoesn't make any difference

It does make a difference: extending the WELCOME message with another field will break existing implementations, and having Options/Details for extensibility without breaking implementations was a major lesson from WAMP v1. So yes, I do care ...

PR #171, opened 2015-08-26
Add Bonefish to the list of WAMP (v2) router implementations.

jpetso:
Putting "yes" in the column for "Advanced Profile" seems presumptious since we don't support the majority of features in the Advanced Profile, but "no" also sounds wrong because some are there (rawsocket, call timeouts). So I put "some" in that column. Hope that works.


oberstet:
@jpetso awesome! done. yes, the info under AP is practically almost useless .. see https://github.com/tavendo/WAMP/issues/172

jpetso:
Thanks!

PR #180, opened 2015-09-01
Revamped meta API to reflect current Crossbar implementation.

ecorm:


ecorm:
For the reviewer's convenience, pre-rendered mardown pages can be viewed here:
- https://github.com/ecorm/WAMP/blob/ecorm-issue-157-redux/spec/advanced/registration-meta-api.md
- https://github.com/ecorm/WAMP/blob/ecorm-issue-157-redux/spec/advanced/session-meta-api.md
- https://github.com/ecorm/WAMP/blob/ecorm-issue-157-redux/spec/advanced/subscription-meta-api.md

Most of the credit goes to @oberstet and Tavendo for implementing the Meta API on crossbar. I just collected all the info together and did some editing. Although much of that "collecting" involved digging through Crossbar's source code. :-)

The formatting of the Meta API reference can be adjusted to your liking later, but at least for now, it's all compiled together for the WAMP community to see.

I didn't know what to do with the old 'Subscriber Meta Events' section having the `SUBSCRIPTION.Options.metatopics` idea, so I left it there. To me, it seems that idea would be best implemented in terms of router meta-events, instead of a special `SUBSCRIPTION.Options` object.

oberstet:
@ecorm Emile, thanks a lot for your work!

ecorm:
Thanks!

PR #193, opened 2015-09-27
Added session statechart design files.

ecorm:
The *.odg files can be opened with LibreOffice draw. You might have to resize the page and add an opaque background for SVG export to work properly. Feel free to do as you please with these files.


oberstet:
coool! thx.

PR #195, opened 2015-10-02
Removed leftover session meta API stuff

ecorm:


goeddea:
@ecorm Thanks for the fix! - Any chance you can find the time to take a look at the RFC draft and give some feedback on the hybi mailing list?

ecorm:
I was planning on doing so this weekend.

PR #213, opened 2015-11-02
Update bp_intro_protocol_overview.md

rogererens:
Connections were introduced twice; should be introduced after Router and Client have been introduced. How about 'transport'?
Some github-bug also changed the 'Wamp advanced profile' section, no idea why (I did not do it)


oberstet:
Thanks!

> Some github-bug also changed the 'Wamp advanced profile' section, no idea why (I did not do it)

yeah, also was wondering and re-reading twice;) Guess it's just whitespace / line endings probably ..

PR #222, opened 2015-12-18
Remove stray inclusions of ticked based auth section

estan:
I believe this should only be included once? (Above, under authentication `ap_authentication.md`)


goeddea:
Correct! Thanks for the fix!

estan:
BTW, I realized after I sent this in that perhaps you appriciate if pull requests include the generated RFC txt/html etc, and not just the .md modification like I did? If so, I'll see about setting up the tooling required.

PR #257, opened 2016-09-22
Fix typo in the RFC

Jopie64:
Before: ... can cancel and issued call ...
After:   ... can cancel an issued call ...

Also some end of line whitespace was removed


goeddea:
Thanks!

Issue #255, opened 2016-09-19
Lack of specification in case of offending messages


angiolep:
What a router shall do if a peer connects and sends any of the offending messages?

- Not deserializable messages (e.g. CIAO)!
- Either SUBSCRIBE, PUBLISH, REGISTER or CALL before any valid HELLO message caused the session open!
- Repeated HELLO messages (please follow https://groups.google.com/forum/#!topic/wampws/J-iBmuHcvpI) within the lifetime of a session!

I cannot read clear indications about the above scenarios on the [WAMP specification](https://tools.ietf.org/html/draft-oberstet-hybi-tavendo-wamp-02)

angiolep:
I had to provide [akka-wamp](https://github.com/angiolep/akka-wamp) (my implementation of the WAMP protocol I wrote for Scala/Akka developers) with the ability to switch between unspecified behaviours :-(

```bash
akka {
  wamp {
     router {
      # The boolean switch to disconnect those peers that send 
      # offending messages (e.g. not deserializable or causing
      # session failures)
      #
      # By default, offending messages are just dropped and 
      # the router resumes processing next incoming messages
      #
      disconnect-offending-peers = false
}}}
```

As described in [akka-wamp/issues/21](https://github.com/angiolep/akka-wamp/issues/21) and [akka-wamp/issues/22](https://github.com/angiolep/akka-wamp/issues/22) the default behaviour I had to decide to the router is to preserve the transport connection (which is a more expensive resource compared to the WAMP session) whatever kind of offending message could provoke a WAMP session failure.

Wouldn't it better if the WAMP protocol defines the behaviour of the router in case of offending messages rather than leaving implementors to guess it (or to provide multiple ones)?

oberstet:
Posting my answer from the mailing list here:

> Yes, we should clarify and augment the text regarding such protocol violations. A router (at least in "production mode") should abriptly disconnect the client upon a protocol violation. There is no point in trying to compensate for a bogus or malicious client. A router may also block the client from further connection attempts. A misbehaving client must be fixed - a router should not expose itself to an extended attack surface resulting from trying to cope with misbehaving clients.

angiolep:
It's good to finally understand that a router shall abruptly disconnect offending clients (those which violate the protocol) in addition to disposing any resources held for them (such as sessions, subscriptions and registrations).

[...]

PR #258, opened 2016-10-03
(#255) Fix lack of specification

angiolep:
This commit explicitly defines what protocol errors are and what
peers shall do when those happen.


oberstet:
Awesome, thanks!

Issue #261, opened 2016-11-08
Message retention (Advanced Profile)


hawkowl:
See https://github.com/crossbario/crossbar/issues/887 for more details, but it seems like we need to define:

- `Publish.Options.retain|bool` (to set it as the topic's retained message, default is False)
- `Subscribe.Options.get_retained|bool` (to request the retained message along with the subscription)
- `Event.Details.retained|bool` (to say that it is a retained message, and not a new message)

And maybe advertise via `Welcome.Roles.broker.message_retention := true`?

cc @oberstet

oberstet:
Yes, feature announcement (`message_retention == true`) from router side is another piece here. I think this essentially completes the spec of this feature.

> Maybe we need an error URI for when the router is allowing a client to publish to the topic, but is not willing to let a particular client publish to a topic with the retain flag set (and the client used acknowledge publication - otherwise it doesn't get the error feedback anyway). But this is a minor thing (and depends on whether we want to let the router admin control this feature at that granularity.

hawkowl:
Yeah, I *think* that's all we'll need. I'll have a look at AB|P and Crossbar, I don't think there'll be any issues implementing it...

I feel like the error URI is unneeded, I can't think of a use case where you'd allow a client to publish to a topic but not change the topic's retain message.

hawkowl:
@oberstet , from the crossbar issue:

> If requested, router should publish the last event on the subscription (we have an object inside the broker already for that) to that particular client (via eligible).

How do you see this working? Would this allow multiple retained messages, for example, in order of the broker receiving them (all with retain = true):

- Message #1: eligible to all
- Message #2: eligible to authid "bar"
- Message #3: eligible to authid "baz"

And then when authid 'baz' connects, they get message #3, authid bar gets #2 on connection, and everyone else gets #1 on connection? And then, if there were a fourth message, eligible to all, it would become the only retained message?

Either that, or there's only one retained message, and clients that aren't eligible just don't get it. I'm not sure, I think the prior option is better, if a little more complicated, as then a single eligible message being retained doesn't make it useless for the rest of the clients.

hawkowl:
I think the former of those two options makes the most sense to a user, so I did up a draft in a PR which defines it: https://github.com/wamp-proto/wamp-proto/pull/262

Haven't written anything like this before so I just tried to copy the style what was there :P

The multiple-retained-messages question only affects the broker, though, so stay tuned for AB|P changes, which we will need first anyway...

[...]

PR #262, opened 2016-11-10
Initial event retention draft

hawkowl:
See https://github.com/wamp-proto/wamp-proto/issues/261

cc @oberstet

### DIFF
--- a/rfc/text/advanced/ap_pubsub_event_retention.md
+++ b/rfc/text/advanced/ap_pubsub_event_retention.md
@@ -1,0 +1,32 @@
+### Event Retention
+
+#### Feature Definition
+
+Event Retention is where a particular topic has an event associated with it which is delivered upon an opting-in client subscribing to the topic.
+
+It can be used for topics that generally have single or very few Publishers notifying Subscribers of a single updating piece of data -- for example, a topic where a sensor publishes changes of temperature & humidity in a data center.
+It may do this every time the data changes (making the time between updates potentially very far apart), which causes an issue fornew Subscribers who may need the last-known value upon connection, rather than waiting an unknown period of time until it is updated.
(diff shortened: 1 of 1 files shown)


hawkowl:
@oberstet I changed it to "Event Retention" as it seems to fit better with the terminology of WAMP, what do you think?

oberstet:
@hawkowl Agreed to both the spec text and the renaming! IMO, this is exactly how this feature should look like.

I will try to raise attention / feedback by posting this on the WAMP mailing list (are you subscribed?): https://groups.google.com/forum/#!forum/wampws

We don't have a formal consensus process for WAMP, but I don't want to just merge it right away either (though I'm completely +1 on this).

meejah:
I think this looks great!

For a concrete use-case, this replaces the pattern of having a "get the state" RPC call paired with an "update the state" PubSub event and collapses it to "publish the state when it changes". It also avoids any question for clients of race-condition type stuff (i.e. do you ask the RPC, and *then* subscribe? if you subscribe and then RPC, what happens if you get a publish before you do the RPC?)

ecorm:
> For a concrete use-case, this replaces the pattern of having a "get the state" RPC call paired with an "update the state" PubSub event and collapses it to "publish the state when it changes".

We are having to do such a get/subscribe pairing in our application. This is a nice, useful feature! :+1:

oberstet:
So we have positive feedback from 3 implementors and no negative feedback. Further, the feature only _adds_ to the AP without changing anything existing (fully optional, fully upwards compat).

Because of above, I am merging=)

oberstet:
@meejah @ecorm yes, this is a nice and common use case! what's also nice: we can add that to WAMP without breaking anything (because, luckily, we've thought about extensibility with WAMP v2 more than with v1;)

PR #264, opened 2016-11-18
Session Resumption

### LINKED ISSUE #260: Disconnect the Session from the Connection
Advanced Profile WAMP brokers may want to implement some form of 'lasting session' for clients. Upon resumption of the session, it will still have all the existing Subscriptions and registrations, although it will not receive messages sent to those subscriptions nor method calls to those registrations that were sent offline -- that is up to the client to fetch using the publication history meta API.

hawkowl:
Fixes #260


hawkowl:
@oberstet so I've taken a stab at this... 

I was thinking, and I am not sure how the multiple HELLO messages would work -- it seems that HELLO/GOODBYE are part of the session, and just implicitly have something to do with realms. Disconnecting from realms and joining as discussed using these two messages would overload them a lot, and I don't think would make too much sense. So, I've sketched it out with two new optional messages -- `JOIN` (and the `JOINED` response) and `LEAVE` (with the `LEFT` response). This completely separates the transport, session, and realm.

Alternatively, I was wondering if we needed to actually disconnect the session and realm... I am trying to think of a use case, maybe you have one I don't know about? Because if we treat a session as being tied to a realm, then we can remove any need for duplicate `HELLO`s or any `JOIN` packets, and implement joining a different realm by starting up a new session on the new realm and connecting to that.

What do you think? I think maybe the latter, unless there's a usecase I don't know about...

oberstet:
Just quickly (need to review the text in detail later):

>  I think maybe the latter ..

Yes, I agree:

* A WAMP session is always tied to a specific realm: one cannot *leave* a session from realm A and *join* that same session to a different realm B or even *rejoin* that same session to realm A again
* In other words: once a WAMP session has *left* a realm, it's gone and void
* No, we should not introduce complete new WAMP messages (I think we have enough extensibility via the options/details fields in HELLO and WELCOME)

I think we also need to be careful when talking about "a session": does it refer to an (abstract) WAMP session thing (like the spec talks about), or does it refer to an implementation artifact (eg `autobahn.wamp.ApplicationSession`).

It may or may not make sense to keep (not destroy) an `ApplicationSession` object even when it has left a realm, and hence there is no abstract WAMP session "existing" anymore.

An `ApplicationSession` object (which is an implementation artifact) at any point in time is either *joined* on a realm or not. It cannot be joined to 2 realms. It can be unjoined, but attached to a transport - in which case it first needs to join a realm to do anything. It can be joined, but not attached to a transport - in which case it is frozen, and (may be) resumable.

hawkowl:
@oberstet so we want to remove the original "The Session can leave the Realm, and join a different Realm, performing a new (different) authentication handshake." ?

oberstet:
@hawkowl Yes. I think I have fallen exactly into the trap I described above: I mixed up talking about `ApplicationSession` (the implementation specific representation of a WAMP session) and the abstract WAMP session.

What I meant with "The Session can leave the Realm, and join a different Realm, performing a new (different) authentication handshake." is the `ApplicationSession` object.

It does not make sense at all to have a WAMP session leave a realm and join a different realm. This would break up the realm separation. Realms ought to be fully separated.

oberstet:
And being able to reuse an `ApplicationSession` object across leaving a realm and joining another one instead of just destructing the object and reconstructing a complete new `ApplicationSession` object seem like a total implementation detail.

oberstet:
@hawkowl also: https://groups.google.com/forum/#!topic/wampws/1Ju4gff1ClE

hawkowl:
Okay, got rid of discussions about sessions leaving realms -- and written it up in an RFC style for inclusion. It's still a little incomplete (need to document the meta events...) and go over it and think of any edge cases, but I think I got them all... still needs polish anyway.

KSDaemon:
Hi all!
Well, i'll try to write all thoughts about this...
First of all, i have a such kind of functionality in [wampy.js.](https://github.com/KSDaemon/wampy.js) and [Loowy](https://github.com/KSDaemon/Loowy). It detects, when websocket connection is aborted, and if it's configured for reconnection (there are _onReconnect_, _onReconnectSuccess_, _autoReconnect_, _reconnectInterval_, _maxRetries_ options), than wampy tries to reconnect, and resubscribe to all previous subscriptions and reregister previously registered RPCs. So, basically, client application may not even notice connection failure.
Proposed specification implies complication and has some gaps:
* I don't feel that separation of session and transport layer really has a big advantage. Right now it's very simple and clean that one session is attached to one transport. That's easy to understand, easy to implement.
* There is no clean specification for how long router should keep session data in cache
* Also this is a potential way for DOS attack: open and abort connections
* How should router process situation, when already there is a connection, and some new client (possibly the same one) is trying to resume session? Should it deny resuming or abort already established one?
* Another bunch of questions: How should router process situation, when client invoke RPC, than abort connection, than callee returns to router a call result? Should router keep invocation result for session? Or throw it away? Same question for subscriptions. Yes, there is event retention feature. But that is something different.

So, at first glance, I think this feature can be implemented at higher level by wamp implementators, without complication of the WAMP specification. May be i'm wrong...

hawkowl:
Hi @KSDaemon :)

So, I'm looking at https://github.com/hawkowl/wamp-proto/blob/f00578d057d3df34a76b00b19bb68196a8ac18f7/rfc/text/advanced/ap_session_resumption.md as the thing that's being proposed right now, I'm not sure if @oberstet wants to extend/change it (they'll hopefully be able to look at this soon), so this is just what I have in my head. It's of course not finished/final yet, so this feedback is good. :)

> I don't feel that separation of session and transport layer really has a big advantage. Right now it's very simple and clean that one session is attached to one transport. That's easy to understand, easy to implement.

I don't think this changes it -- clients that want resumable sessions are 100% compatible with routers that don't wish to implement this. To actually get 100% reliable reconnected sessions, you will have to implement some sort of client reconnection that can handle the session going away (for example, if the router doesn't have persistent storage for it, which Crossbar will 99% likely not have, for example)

> There is no clean specification for how long router should keep session data in cache

This is on purpose and in-line with similar protocols (e.g. MQTT). MQTT actually specifies that a session lasts as long as the router does, but the ability for a WAMP router to decline resuming an otherwise valid session leaves this up to the administrators.

> Also this is a potential way for DOS attack: open and abort connections

Open and freeze sessions? You'd have to authenticate to the router to make a valid session, so handling this is the same as any other abuse prevention. And because of the best-effort basis, the router can always choose to drop sessions.

> How should router process situation, when already there is a connection, and some new client (possibly the same one) is trying to resume session? Should it deny resuming or abort already established one?

In the RFC entry I have written:

> If a Client attempts to resume a non-paused Session on the Broker, the Broker MUST detatch the Session from the Client it is currently attached to, and attach it on the new Client. It must detatch it by way of a GOODBYE message with Goodbye.Details.resumable|bool set to false (as the single-use token will have been used by the new client) and Goodbye.Reason|uri set to wamp.error.other_client_attached. This can be used for Sessions which are attached to Clients that are disconnected, but have not yet timed out (for example, on high-latency wireless networks).

I believe this is the only valid way to do it, as high latency connections (e.g. 3G) may leave a session that isn't timed out under TCP for a long while, even when the client returns.

 > Another bunch of questions: How should router process situation, when client invoke RPC, than abort connection, than callee returns to router a call result? Should router keep invocation result for session? Or throw it away? Same question for subscriptions. Yes, there is event retention feature. But that is something different.

Ooh, so, RPC is an interesting edge case. As far as pubsub, I think event retention (so a slightly smarter client) might be what we're going with, but buffering is an option. I know @oberstet had concerns about sessions buffering being a really big RAM hog and potential DoS vector, but I personally think that it may be a valid option (considering that in the MQTT spec, it is mandated that servers buffer all subscriptions to disconnected sessions for as long as the session is active [which is, again, for as long as the router is up]) that we can write affordances into the spec to allow router administrators to purge large sessions, or limit the buffer (and allow the client to use event history for the missing items). I'm not entirely sure what direction @oberstet wants to go in this area, and I think there's good arguments for buffering vs event retention only.

The RPC call is an interesting edge case -- I believe this would tie into QoS stuff (which is also on the radar for me, but only makes sense when you have session resumption). That is, if you send an ack'd publish from the server and the client never responds and times out later, that on session resumption, it sends publishes/events the router sent *before* it detached the session but never got a response to (with a duplicate flag, I guess), even if there is no buffering of events while it was detached. Again, this only makes sense when we have resumable sessions, so I have not yet done any spec work on that.

Thanks for the feedback though, it's important! I'll have a think over the issues you've raised and how I can fix them/make their solutions in the RFC clearer, and when @oberstet is back from confs we can see how we'll move this forward.

oberstet:
@KSDaemon I think there are multiple features being discussed here: client initiated automatic resubscription/reregistration implemented in the WAMP client library versus router side session freeze and resume.

The latter actually queues events (and maybe even calls) for the specific session. This will consume resources on the router side, and hence there should be knobs to control the final timeout and destruction of frozen session for the router administrator. But this is an implementation thing and should be in the spec. Yes, you are right, this isn't the same as "event retention", and it is also not the same as "event history".

With the latter, the client can retrieve a certain amount of historical events (and hence do QoS "at least once" event processing), but the Client needs to know what to request. With "session resumption", the Router needs to keep and remember the events per Client session that is maintained - but the upside on the Client side is: the Client can catch up *exactly* on the events it missed (and hence do QoS "exactly once" event processing).

oberstet:
Rgd the specific points raised by @KSDaemon :

> I don't feel that separation of session and transport layer really has a big advantage. Right now it's very simple and clean that one session is attached to one transport. That's easy to understand, easy to implement.

There is an aspect in the spec that many (well, AutobahnJS for sure) client libraries don't expose: a session can leave, and then, without closing the transport, a new session can be established over the existing transport.

I only mention this, because originally thought as you, that the spec would be easy to grasp, but the ability to have multiple different sessions running over one transport one after the other complicates implementation in my experience.

So I'd say, this is a chance to come up with something more flexible and even clearer.
 
> There is no clean specification for how long router should keep session data in cache

We should mention that this is an implementation decision how to expose knobs to control the resource consumption and reclaiming.

> Also this is a potential way for DOS attack: open and abort connections

There are many DoS vectors already, like with 5s WebSocket opening handshake timeout. I think this cannot be actually solved in a WAMP router, but the best a Router could do would be to detect malicious Clients, and send the IPs to an embedded firewall that then blocks further connections coming from that IP. An embedded firewall (running a kernel module or even running filters on network hardware) can filter out traffic coming from malicious Clients.

> How should router process situation, when already there is a connection, and some new client (possibly the same one) is trying to resume session? Should it deny resuming or abort already established one?

This is a good point: I could think of 2 behaviors here: a) the last connecting session wins (and the other gets kicked automatically), or b) the new session gets an error (GOODBYE) because there already is another transport with a resumed session already. Maybe a knob?

> Another bunch of questions: How should router process situation, when client invoke RPC, than abort connection, than callee returns to router a call result? Should router keep invocation result for session? Or throw it away? Same question for subscriptions. Yes, there is event retention feature. But that is something different.

The question of whether the proposed feature of queuing events for sessions is something that can be transferred to RPC is definitely interesting! But I think we should do that separately, as I can imagine this will need some new ideas and longer discussions. Whereas the idea of the feature for PubSub is quite straight forward: queue up events the session that a session missed and redeliver to the session when it resumes.

oberstet:
I think the subject matter here is a pretty important one, or one of the few last more complex areas where we need to work out features/semantics in WAMP. I don't have the time/patience right now to dive into - there are lots of aspects. This is also one where we definitely need a broader discussion with other implementors. Backlog ..

gammazero:
There are some significant architectural ramifications that depend on the intent of this feature proposal.  If the basic intent of this feature is to tolerate temporary connectivity loss, then this is really a router feature, not a broker or dealer feature.

Session establishment is not a broker or a dealer concern, so session resumption should not be either.  The PR states that the feature is supported by the router, not the broker or dealer role.  So, I believe the specification for this feature needs to work from the premise that broker and dealer states are not affected (do not see) a loss and resumption of a session.

@oberstet This means that queuing RPC messages cannot be considered separately, since the broker and dealer would have no knowledge of connection loss and resumption (otherwise results in a layer violation).

mbonneau:
Adding a Session Resumption feature to compensate for unreliable connections is not an easy task and I feel is too much to try and add to the spec.

WAMP is a stateful protocol. If the client and the router are out of sync at all, then the state of the session is invalid.

Consider a client sending a `REGISTER` message and then losing its network connection. The router may send back a `REGISTERED` message and build up the registration while the transport is invalid.  The client then comes back and resumes the session but as far as the client knows, it is still waiting for a response to the `REGISTER` request. The client may even receive invocations for a registration id that it is unfamiliar with. (similar issues with `CALL`, `INVOKE`, `PUBLISH`, `SUBSCRIBE`, etc.)

In order to overcome this, you would need a session-synchronization algorithm which would be complicated to add to the spec.

It still may be useful to add Session Resumption for clients that wish to step away for a moment (like switching apps on your phone). In that case there can be a clear agreement between the router and client of when state is frozen and maybe messages could be queued for a time.

oberstet:
just very quickly, my view:

* this is a very complex feature, and we should be really careful, be really certain what and why we wanna have sth like this
* what a router can could is this: queue up events being dispatched to a subscriber while that is offline, queue up calls being dispatched to a callee while that is offline

handling breakage of a session in the middle of a REGISTER/REGISTERD: huh, no. I think and agree this is a road to .. insanity;)

oberstet:
so maybe one way to advance the discussion might be to contrast this feature with that proposal https://github.com/wamp-proto/wamp-proto/issues/293

note, that we might come up with any combination of this, the other or neither features .. I guess we should be careful here, not adding unnecessary complexity - in as, what was the use case exactly again? if we have a clearly identified and defined use case, we can asses the importance of the former, and then can weigh the added complexity against ...

gammazero:
I vote to close this PR.  This PR has been here for a long time, and there is no consensus on whether this belongs in the spec.  

In the interest of not accumulating complication in the spec, if a proposed feature does not have a clear reason _and_ a clear solution (as agree by majority of people here), then we should drop it.  Many  good ideas are better left to individual implementations.

oberstet:
I agree closing this now - but save the work done (the text) somewhere, so we _might_ come back or only reference later.

PR #268, opened 2016-11-22
Last Will/Testament

### LINKED ISSUE #266: Last Will/Testament for WAMP sessions/transports
See https://github.com/crossbario/crossbar/issues/886 for discussion.

From @oberstet :

If we'd use the WAMP/CB meta API, we could have:

    wamp.session.publish_testament(<topic>, <args>, <kwargs>, <publish_options>, <scope: "transport" | "session")
    wamp.session.flush_testament(<scope: "transport" | "session">)

Which would allow submitting multiple last wills (that later would be published as originally submitted), and to flush (remove) any previously submitted last wills.

hawkowl:
Here's a draft... cc @oberstet 

Fixes #266

### DIFF
--- a/rfc/draft-oberstet-hybi-tavendo-wamp.md
+++ b/rfc/draft-oberstet-hybi-tavendo-wamp.md
@@ -214,6 +214,8 @@ While implementations MUST implement the subset of the Basic Profile necessary f
 
 {{rfc/text/advanced/ap_pubsub_topic_reflection.md}}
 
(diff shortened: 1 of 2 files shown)


oberstet:
The feature as described in the PR is implemented in at least Crossbar.io / Autobahn. I think it's semantics are pretty straight forward. I am going to merge this, despite there hasn't been much feedback from other implementors. I am doing this in the hope that we do get feedback once it is in the spec - this isn't set in stone, so when others disagree, we can still change things later.

PR #269, opened 2017-03-19
Fix typo in protocol word

pahaz:


oberstet:
Thanks! btw, not sure if you are aware, all files but the `rfc/text/basic/bp_intro_design_philosophy.md` file are actually generated `https://github.com/wamp-proto/wamp-proto/blob/master/Makefile#L3` .. we just commit these built files to the repo too for convenience

PR #278, opened 2017-04-04
Fix improper “it’s” in several spots.

FGasper:


oberstet:
Awesome, thanks for going all over the text and correcting.

PR #290, opened 2018-02-12
Described call canceling. closes #140. closes #141.

### LINKED ISSUE #140: More info about canceling rpc call options
Right now in advanced profile there is only one string, about posible options:
CANCEL.Options.mode|string == "skip" | "kill" | "killnowait"

We should extend docs with explanation, what do this options mean, and how they should be done on dealer/callee side.

### LINKED ISSUE #141: What if the Callee does not support CANCEL?
If a Caller attempts to cancel an RPC where the Callee does not support CANCEL, what should the behavior be?

Should the router send back an error?
Should the router send back the eventual results even though the Caller has requested the cancelation?
What error URI should be sent back to the Caller?

I would suggest:

The router send back `wamp.error.not_supported`.

At that point the call is still not canceled and should behave the same as if the CANCEL was never issued.

This behavior could be overridden by sending an option `CANCEL.Options.discard_results|bool := true` at which point the router will discard the results when they are sent back from the Callee. A `wamp.error.canceled` will be sent back to the Caller.

KSDaemon:
Guys, i've updated spec, based on our discussions. Please have a look, and may be even merge if it is ok :)
I think we should always use Pull requests with minimum one man review for updating spec.

### DIFF
--- a/rfc/text/advanced/ap_rpc_call_canceling.md
+++ b/rfc/text/advanced/ap_rpc_call_canceling.md
@@ -8,8 +8,17 @@ The message flow between *Callers*, a *Dealer* and *Callees* for canceling remot
 
  * `CANCEL`
  * `INTERRUPT`
+ * `ERROR`
 
-A call may be cancelled at the *Callee*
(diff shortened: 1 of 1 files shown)


oberstet:
looks good! merging ..

PR #294, opened 2018-02-20
Explicitly state format of signature string

gammazero:
Example shows what appears to be a base64-encoded string in the CRA response, but the format needs to be explicitly stated.

### DIFF
--- a/rfc/text/advanced/ap_authentication_cra.md
+++ b/rfc/text/advanced/ap_authentication_cra.md
@@ -53,7 +53,7 @@ The client needs to compute the signature as follows:
 
 That is, compute the HMAC-SHA256 using the shared `secret` over the `challenge`.
 
-After computing the signature, the client will send an `AUTHENTICATE` message containing the signature:
+After computing the signature, the client will send an `AUTHENTICATE` message containing the signature, as a base64-encoded string:
 
 {align="left"}
 ```javascript


bbdb68:
Hi

I am trying to implement WAMP CRA (wihtout salting) and I am experiencing difficulties

* should the secret be also base64 encoded ? sound like yes, but I did not see it in the spec
* so what about the challenge ? should we use it "as is" ?
* hmac[256] returns a 64 bit long signature, but looking at authobahn cpp, it seems taht only the first 32 bytes are used ?

my current understanding is 
*base64 encore the scecret
* compute hmac256 with the base 64 encoded secret and the raw challenge
* base64 encore the 32 first bytes of the hmac output

Is this correct ?

gammazero:
@bbdb68
- It is not required that the secret be encoded in any particular way, as long as the client and the server both use the same string.  So, that is really up to what/how you store and retrieve secrets.  That is what the spec does not specify
- The challenge should be used "as is".  That is, whatever json string the server sends it exactly what the client should use as the challenge, since that must be what the server used as the challenge.
- I have to look at other implementations, but I did not see that only the first 32 bytes of signature were used.  I know that 32 bytes is the default key length for a pbkdf2 derived key, but as far as I know for CRA the whole signature is compared.

bbdb68:
Thanks for these informations.

from authobahn.cpp, in the compute_wcs function of
https://github.com/crossbario/autobahn-cpp/blob/master/autobahn/wamp_auth_utils.hpp,
I understood that because of the size of hash array, only 32 bytes were
used.

Do you know where I could find a  valid sample of (secret, challenge,
signature), in order to write a unit test ?

Bruno

PR #300, opened 2018-03-13
WAMP-SCRAM

### LINKED ISSUE #135: WAMP-SCRAM specification
As discussed in https://github.com/tavendo/WAMP/issues/128 _WAMP-CRA_ is unsafe in regards to database thefts. A new and better authentication method is required. [SCRAM](http://en.wikipedia.org/wiki/Salted_Challenge_Response_Authentication_Mechanism) turned out to be very promising.

[oberstet](https://github.com/oberstet) already gave a summary (https://github.com/tavendo/WAMP/issues/128#issuecomment-74271094):

> From looking at SCRAM, this claims to provide protection against these attack vectors:
> 
> Eavesdropping
> Replay
> Database Compromise
> and also against dictionary attacks and password collisions.
> 
> It additionally claims to protect against impersonating a server (after the original server had signed up a user).
> 
> Here are some references:
> 
> http://en.wikipedia.org/wiki/Salted_Challenge_Response_Authentication_Mechanism
> https://tools.ietf.org/html/rfc5802
> http://www.mongodb.com/blog/post/improved-password-based-authentication-mongodb-30-scram-explained-part-1
> http://www.mongodb.com/blog/post/improved-password-based-authentication-mongodb-30-scram-explained-part-2
> The good news is, with the WAMP messages HELLO, CHALLENGE, AUTHENTICATE and WELCOME (and the extensibility built in those messages), we have the required message flow in place without changing WAMP.
> 
> However, I don't think it makes sense to put this into WAMP-CRA as it exists.
> 
> I think we should come up with a mapping of SCRAM to WAMP: WAMP-SCRAM (wamp-scram)
> 
> This should be pretty straight-foward, and we can reuse an already designed, scrutenized scheme.
> 
> We also do not need new crypto primitives, as the following will do fine:
> 
> SHA256
> HMAC
> PBKDF2
> XOR

ecorm:
Resolves #135.


ecorm:
Will submit a similar proposal soon using a PAKE protocol (e.g. SRP) instead of SCRAM. Should wait before merging this, but feedback is still welcome.

rfc/text/advanced/ap_authentication_scram.md
```diff
@@ -0,0 +1,387 @@
...
+Username and password strings SHALL be normalized according to the _SASLprep_ profile described in [RFC4013](https://tools.ietf.org/html/rfc4013), using the _stringprep_ algorithm described in [RFC3454](https://tools.ietf.org/html/rfc3454).
+
+While SASLprep preserves the case of usernames, the server MAY choose to perform case insensitve comparisons when searching for a username in the authentication database.
+
+
+##### Channel Bindings
+
+_Channel binding_ is a feature that allows higher layers to establish that the other end of an underlying secure channel is bound to its higher layer counterpart. See [RFC5056](https://tools.ietf.org/html/rfc5056) (_On the Use of Channel Bindings_) for an in-depth discussion.
+
+[RFC5929](https://tools.ietf.org/html/rfc5929) defines binding types for use with TLS transports, of which `"tls-server-end-point"` may be used with SCRAM. Note that [RFC7677, section 1](https://tools.ietf.org/html/rfc7677#section-1), states that TLS did not have the expected properties for `"tls-unique"` to be secure.
```
oberstet: Mmh. "did not have the expected properties" for what? =) seriously, here is how I read it. the RFC7627 introduces a "Extended Master Secret Extension" which when used makes the master secret, and thereby the session hash, cover _also_ the peer certificate. and then, SASL would be secure when using tls-unique channel binding. IOW: we would say SHOULD use channel binding and SHOULD use extended master secret extension. With tls-server-end-point: as I understand that, the channel is "only" bound to the public bits of the peer certificate in this case - or the private bits too?

rfc/text/advanced/ap_authentication_scram.md
```diff
@@ -0,0 +1,387 @@
+#### Salted Challenge Response Authentication Mechanism {#wamp-scram}
+
+The WAMP Salted Challenge Response Authentication Mechanism ("WAMP-SCRAM"), is a password-based authentication method where the shared secret is neither transmitted nor stored as cleartext. WAMP-SCRAM is based on [RFC5802](https://tools.ietf.org/html/rfc5802) (_Salted Challenge Response Authentication Mechanism_) and [RFC7677](https://tools.ietf.org/html/rfc7677) (_SCRAM-SHA-256 and SCRAM-SHA-256-PLUS_).
+
+WAMP-SCRAM supports the Argon2 ([draft-irtf-cfrg-argon2](https://datatracker.ietf.org/doc/draft-irtf-cfrg-argon2/)) password-based key derivation function, a memory-hard algorithm intended to resist cracking on GPU hardware. PBKDF2 ([RFC2898](https://tools.ietf.org/html/rfc2898)) is also supported for applications that are required to use primitives currently approved by cryptographic standards.
+
+##### Security Considerations
+
```
oberstet: this is awesome!!

rfc/text/advanced/ap_authentication_scram.md
```diff
@@ -0,0 +1,387 @@
...
+##### Conventions
+
+###### Base64 encoding
+
+Base64 encoding of octet strings is restricted to canonical form with no whitespace, as defined in [RFC4648](https://tools.ietf.org/html/rfc4648) (_The Base16, Base32, and Base64 Data Encodings_).
+
+###### Nonces
+
+In WAMP-SCRAM, a _nonce_ (number used once) is a base64-encoded sequence of random octets. It SHOULD be of sufficient length to make a replay attack unfeasible.
```
oberstet: could we add a recommendation (a default for implementations to use)? otherwise I need to dig into another RFC;) seriously, we should recommend a high secure, but practical value ..

rfc/text/advanced/ap_authentication_scram.md
```diff
@@ -0,0 +1,387 @@
...
+_Channel binding_ is a feature that allows higher layers to establish that the other end of an underlying secure channel is bound to its higher layer counterpart. See [RFC5056](https://tools.ietf.org/html/rfc5056) (_On the Use of Channel Bindings_) for an in-depth discussion.
+
+[RFC5929](https://tools.ietf.org/html/rfc5929) defines binding types for use with TLS transports, of which `"tls-server-end-point"` may be used with SCRAM. Note that [RFC7677, section 1](https://tools.ietf.org/html/rfc7677#section-1), states that TLS did not have the expected properties for `"tls-unique"` to be secure.
+
+
+##### Authentication Exchange
+
+WAMP-SCRAM uses a single round of challenge/response pairs after the client authentication request and before the authentication outcome.
+
+The mapping of RFC5802 messages to WAMP messages is as follows:
```
oberstet: this is also perfect! very simple and straightforward ..

rfc/text/advanced/ap_authentication_scram.md
```diff
@@ -0,0 +1,387 @@
...
+            "cbind_data": null
+        }
+    ]
+```
+where:
+
+1. `Signature|string` argument: The base64-encoded `ClientProof`, computed as described in the [SCRAM-Algorithms](#scram-algorithms) section.
+2. `nonce|string`: The concatenated client-server nonce from the previous `CHALLENGE` message.
+3. `channel_binding|string`: Optional string containing the channel binding type that was sent in the original `HELLO` message.
+4. `cbind_data|string`: Optional base64-encoded channel binding data. MUST be present if and only if `channel_binding` is not `null`. The format of the binding data is dependent on the binding type. For `tls-server-end-point`, this is a hash of the TLS server's certificate, computed as described in [RFC5929, section 4.1](https://tools.ietf.org/html/rfc5929#section-4.1).
```
oberstet: no, this isn't sth required for `tls-unqiue`, and that is the channel binding I would like to use ..

rfc/text/advanced/ap_authentication_scram.md
```diff
@@ -0,0 +1,387 @@
...
+{align="left"}
+```javascript
+    [2, 3251278072152162,
+        {
+            "authid": "user",
+            "authrole": "frontend",
+            "authmethod": "wamp-scram",
+            "authprovider": "static",
+            "roles": ...,
+            "authextra":
```
oberstet: yes, if the wamp-scram here would be part of a future multi-factor scheme, then we would need a list of dicts for WELCOME.authextra here

rfc/text/advanced/ap_authentication_scram.md
```diff
@@ -0,0 +1,387 @@
...
+SCRAM uses a password-based key derivation function (KDF) to hash user passwords. WAMP-SCRAM supports both [Argon2](https://datatracker.ietf.org/doc/draft-irtf-cfrg-argon2/) and [PBKDF2](https://tools.ietf.org/html/rfc2898) as the KDF. Argon2 is recommended because of its memory hardness and resistance against GPU hardware. PBKDF2, which does not feature memory hardness, is also supported for applications that are required to use primitives currently approved by cryptographic standards.
+
+The following table maps the `CHALLENGE.Details.kdf` type string to the corresponding KDF.
+
+KDF type string | Function
+--------------- | --------
+`"argon2id13"` | Argon2id variant of Argon2, version 1.3
+`"pbkdf2"`      | PBKDF2
+
+WAMP-SCRAM client implementations SHALL support **all** of the above KDFs. During authentication, there is no "negotiation" of the KDF, and the client MUST use the same KDF than the one used to create the keys stored in the authentication database.
```
oberstet: and servers? I guess both client and router implementations SHOULD support all KDFs? plus: ".. and SHOULD prefer argon2id13 to pbkdf2"?

oberstet:
fantastic job! really great level of polish. please see my comments in the review above. all minor stuff .. maybe the one I have concerns with is the channel binding. I'm not sure, but this is virtually the only thing. So, modulo that, I am +1 on merging!

rfc/text/advanced/ap_authentication_scram.md
meejah: Isn't it more important for servers to support "all" KDFs? Then client implementations can choose to support only one (i.e. could be "SHOULD support all KDFs" for clients and "MUST support all KDFs" for servers).

rfc/text/advanced/ap_authentication_scram.md
meejah: Maybe just add, "at least 16 bytes of random binary data"..?

rfc/text/advanced/ap_authentication_scram.md
```diff
@@ -0,0 +1,387 @@
...
+
+{align="left"}
+```javascript
+    [4, "wamp-scram",
+        {
+            "nonce": "egVDf3DMJh0=SBmkFIh7sSo=",
+            "salt": "aBc+fx0NAVA=",
+            "kdf": "pbkdf2",
+            "iterations": 4096,
+            "memory": null
```
meejah: Maybe just name this "cost" so it can be consistent between `pbkdf2` and `argon2id`?

rfc/text/advanced/ap_authentication_scram.md
```diff
@@ -0,0 +1,387 @@
...
+##### Key Derivation Functions
+
+SCRAM uses a password-based key derivation function (KDF) to hash user passwords. WAMP-SCRAM supports both [Argon2](https://datatracker.ietf.org/doc/draft-irtf-cfrg-argon2/) and [PBKDF2](https://tools.ietf.org/html/rfc2898) as the KDF. Argon2 is recommended because of its memory hardness and resistance against GPU hardware. PBKDF2, which does not feature memory hardness, is also supported for applications that are required to use primitives currently approved by cryptographic standards.
+
+The following table maps the `CHALLENGE.Details.kdf` type string to the corresponding KDF.
+
+KDF type string | Function
+--------------- | --------
+`"argon2id13"` | Argon2id variant of Argon2, version 1.3
+`"pbkdf2"`      | PBKDF2
```
meejah: Is there a version 1.9 or similar? The python library `argon2_cffi` refers to its latest argon2id hash as "v=19" is why I ask..

rfc/text/advanced/ap_authentication_scram.md
ecorm: I'll try to find an authoritative source for a suggested length, other than "some guy on Stack Exchange".

rfc/text/advanced/ap_authentication_scram.md
```diff
@@ -0,0 +1,387 @@
...
+```javascript
+    [1, "realm1",
+        {
+            "roles": ...,
+            "authmethods": ["wamp-scram"],
+            "authid": "user",
+            "authextra":
+                {
+                    "nonce": "egVDf3DMJh0=",
+                    "channel_binding": null
```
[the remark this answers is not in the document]
ecorm: I think we should proceed as if the scheme proposed in #301 doesn't exist. If/when the 301 scheme is adopted, we can revise **all** authmethods accordingly.

rfc/text/advanced/ap_authentication_scram.md
ecorm: You have to keep in mind that *user registration* is a whole other process that is separate from *user authentication*. After the user is registered and the hashed keys are stored in the authentication database, the client no longer has any choice whether to use argon2id13 or pbkdf2 for that particular user. The client *must* authenticate using the same hash algorithm that was used for user registration.

For example, one router can choose to only store argon2id13-derived keys, and another router chooses to store only pbkdf2-derived keys. For a WAMP client implementation to universally support all routers, the client implementation would need to support both argon2id13 and pbkdf2.

If user registration was part of the spec, then I could see how it would be desirable to require that routers support both argon2id13 and pbkdf2. But since we don't impose any requirements for user registration, I don't see the need to impose both argon2id13 and pbkdf2 for router implementations.

rfc/text/advanced/ap_authentication_scram.md
ecorm: Both `pbkdf2` and `argon2id` refer to this parameter as "iterations". It was bcrypt that was different, but we've now agreed to remove bcrypt as a possible KDF.

meejah:
Looks great overall. I left a few inline comments. I don't know of a good source for "how long should the nonces be". They need to be long enough so that the chances the same client uses the same nonce again (ever) is "negligible".

Using 16 bytes means `8**16)` so after half that number of authentications, a collision is expected -- so, seems safe? (I mean, we could probably get away with less .. but)

rfc/text/advanced/ap_authentication_scram.md
ecorm: As I've mentioned already in another related issue somewhere, 19 is the decimal representation of the hexadecimal number 0x13. Argon2 has chosen to use hexadecimal numbers to represent its version number.

rfc/text/advanced/ap_authentication_scram.md
meejah: I see -- so you're imposing the "SHALL" on client implementations so that interop is more likely -- but isn't it just as likely that some client decides "we'll only do `argon2id13`" as that some router will decide that? I mean, I could see that being "policy" for an application just as easily as a router...?

"The thing that's doing registrations" would be where the KDF is negotiated (if at all), so you'd just fail to be able to register if you were using a router that said "`argon2id13`-only" and your client only did `pbkdf2` (or vice versa obviously).

Also worth noting that I presume the registration-process would involve the server/router telling the client: KDF, cost, memory, salt and then the client hashing its password and sending the appropriate results back to the server -- maybe it's worth mentioning this briefly somewhere in the spec?

rfc/text/advanced/ap_authentication_scram.md
meejah: ah, okay then nevermind :)

rfc/text/advanced/ap_authentication_scram.md
meejah: Ahh, sorry I missed that ;)

ecorm:
Using the [square approximation of the birthday problem](https://en.wikipedia.org/wiki/Birthday_problem#Square_approximation), if the random nonce is 64 bits, with 2^16 = 65536 tries, the approximate probability of collision would be `p ~= N^2 / 2*M = (2^16)^2 / 2*2^64 = 2^32 / 2^65 = 1 / 2^33`, so 1 in 8 billion. With 2^32 tries, the probability becomes `2^64/ 2^65 = 1/2 = 50%`.

If the random nonce is increased to 128 bits, the probably becomes 1/1.6e29 for 65536 tries, and 1/3.7e19 for 2^32 (over 4 billion) tries.

To summarize:

nonce bit length | tries | probability of collision
--------------------- | ------ | ----------------------------
64 | 65535 | 1 in 8.2 billion
64 | 4.1 billion | 1 in 2
128 | 65535 | 1 in 1.6*10^29
128 | 4.1 billion | 1 in 3.7*10^19

If those calculations are correct, then I would be comfortable recommending that the nonce SHOULD be at least 128 bits.

meejah:
Ah, yeah birthday problem will apply here, won't it. Okay, 32 bytes then (maybe some of these implementations use 16 bytes because there's a server and client nonce..? but I don't want to have to analyze if or why that matters ;) )

ecorm:
Similar [formula here in this SE answer](https://security.stackexchange.com/a/6157/169835), expressed in powers of two. So, `nonce_bits = 2*tries_bits + improbability_bits`. For 4.1 billion tries and collision probably of one in 4.1 billion, `nonce_bits = 2*32 + 32 = 96`. Might as well tack on another 32 bits to make 128 bits.

> Okay, 32 bytes then

16 bytes for client nonce and 16 bytes for the server nonce, which results in a 32 byte combined nonce. The attacker can just reuse the same nonce, so we should rely on only the victim peer to provide the randomness.

rfc/text/advanced/ap_authentication_scram.md
ecorm: I now realize that requiring the client to support both algorithms, but not the router, is going to be confusing to implementors/adopters, so I'm willing to change it so that both client/server are required to support both algorithms. If there's an application/site policy of only using a certain KDF, then they can configure their router accordingly.

I someone wants to later change the spec it so that it's not required to support both algorithms, they can raise an issue and make their case.

> Also worth noting that I presume the registration-process would involve the server/router telling the client: KDF, cost, memory, salt and then the client hashing its password and sending the appropriate results back to the server -- maybe it's worth mentioning this briefly somewhere in the spec?

The registration process does not necessarily involve the client and server talking to each other. The router administrator could directly add the user and assign a known password by configuring a file or manipulating a database.

If a router implementation decides to provide a client API for registering a user, the hashing could be done on the server side if the cleartext password is sent directly by the client (presumably over a secure encrypted transport). Alternatively, the hashing could be done client-side if the server first sends it the algorithm/salt/cost parameters. There are so many possible permutations that I'm not comfortable specifying exactly how it should be done.

Given that the server does not need the cleartext password, it would be preferable from a security standpoint to have it hashed client side during registration that involves. But that would place a burden on the registration client on knowing how to compute the hashes and stored keys.

meejah: I don't see the point of using SCRAM if the server/router ever sees the plaintext password, personally. The client already knows how to compute the correct pieces, since it has to do that to authenticate in the first place (i.e. it has to produce `SaltedPassword` and `StoredKey` already; "registration" would merely be sending those along to the server with the parameters used).

Anyway, that said I agree with you that there's lots of ways to do it so maybe suggesting any at all is counter-productive. "Outside the scope" is fine :)

A very simple registration method would have the client produce the values and send them all to the server -- the server could reject it if e.g. the parameters were too simple, etc. (The server could compute `ServerKey` itself from `SaltedPassword`). So, yes, again this doesn't involve the server talking "first" and sending params.

oberstet: I am +1 on: "both clients and routers SHOULD implement both KDF functions". there is no point in "MUST" - I am simply going to ignore that as an implementor and can claim the app code just disabled or doesn't use the respective unimplemente KDF;) We should leave that to implementors - as with implementing WAMP-SCRAM at all ..

oberstet:
FWIW, WAMP-cryptosign uses 32 bytes random challenges (https://github.com/crossbario/crossbar/blob/master/crossbar/router/auth/cryptosign.py#L86) where the challenge sent to the client must be XORed with the channel ID by the authenticating client first, and then signed and sent back to the router.

rfc/text/advanced/ap_authentication_scram.md
ecorm: Hmm... at first you guys wanted both to be mandatory to avoid negotiating the KDF, and now you want it to be up to the implementation. Is this change of mind due to better understanding that there's no KDF negotiation involved during user authentication, but only possibly during user registration?

I'm fine with making it "both clients and routers SHOULD implement both KDF functions", as long as we cease flip-flopping on this matter. :smile:

oberstet: right;) trying to nalil it, unify: both KDFs should be possible, but IMO we should take an opinionated stand here in favor of argon:

* argon: both client/router implementations MUST support it
* pbkdf2: both client/router implementations SHOULD support it

so this leaves open the option to run pbkdf2 in environments where this is necessary or wanted for whatever reasons, but allows implementations to only implement argon, and makes it the default choice

oberstet:
@ecorm again, thanks a lot for this awesome new part in the spec! unfortunately, this makes the parts on the other authentication methods (if there are parts at all) look quite old =)

from my side, adding a notice "channel binding under discussion" with the link to a new follow up github issue would be cool and good enough, as the channel binding thing should not block merging this PR.

rgd implementation support for KDF (the MUST vs SHOULD etc discussion above): chose as you like;)

let me know when you want it merged, or merge it (as you have write permission).

ecorm:
> Rgd channel binding, it would probably make sense to not only include client and server chosen nonces (which is good), but also the channel ID when being challenged. Note that the channel ID must NOT be sent over the wire as eg another attribute (like nonce), but the peer implementation should prove that its able to include the TLS channel ID when computing the challenge/signature at the WAMP session level. This again ties the WAMP session to the underlying TLS session, so that the former cannot be peeled off the latter.

I'm not familiar with what you call "TLS channel ID" and will have to do some research. Perhaps it's already part of what the SCRAM RFC calls "channel binding data".

The SCRAM RFC requires that a hash of the client's "channel binding data" be sent over the wire. The channel binding data is not sent in cleartext. The hash of the channel binding data is also mixed into the `AuthMessage`, along with the client-server nonces, that is used to compute the `ClientProof` and `ServerSignature`.

We have to be extremely careful about adding or changing things to the algorithms specified in the SCRAM RFC, or risk moving into the territory of "home brew" cryptography, with all the perils that entails.

ecorm:
@oberstet Thank you for the kind words. Before merging, I want to add a recommendation of at least 128-bits for the random nonce, as well as changing the KDF support to SHOULD for both algorithms and both client/server.

I also want to do a little bit more research on TLS channel binding to make sure I got it right in this spec. The SCRAM RFC is rather agnostic about channel binding, since it's specific to the transport being used.

ecorm:
> Does the spec require that the server checks that the hash of the channel binding data (the "ID" of the channel) is equal to the one the server sees on his side of the TLS connection?

Yes it does. I admit that this should be clearer in my draft. Right now, under 
[Final Client Authentication Message](https://github.com/ecorm/wamp-proto/blob/ecorm-scram/rfc/text/advanced/ap_authentication_scram.md#final-client-authentication-message), it says:

> The `cbind_data`, if applicable, is valid.

I'll add more text explaining that the `cbind_data` must match what the server sees on its side of the connection.

oberstet:
> I'll add more text explaining that the cbind_data must match what the server sees on its side of the connection.

Yeah, because at least as far as I understand, this is crucial for the channel binding to be effective at all.

The other question I have: why send the hash over the wire at all, if both sides have access to the same source material that goes into the hash anyways?

meejah:
I thought it worked like: the client sends e.g. "p=tls-unique" and the server responds with the actual channel-binding data? Or are you saying the client also includes "c=<channel-binding-data>" in that message as well..?

ecorm:
@oberstet 
> The other question I have: why send the hash over the wire at all, if both sides have access to the same source material that goes into the hash anyways?

I don't know why. Perhaps they wanted to it leave flexible for other non-TLS transports? [RFC5929, section 10.2, On Disclosure of Channel Bindings Data by Authentication Mechanisms](https://tools.ietf.org/html/rfc5929#section-10.2), addresses this issue for TLS channel bindings.

@meejah 
> I thought it worked like: the client sends e.g. "p=tls-unique" and the server responds with the actual channel-binding data? Or are you saying the client also includes "c=" in that message as well..?

No, the server doesn't send any channel binding information.

Here's the SCRAM RFC example exchange, modified for `p=tls-unique` channel binding:

```
   C: p=tls-unique,,n=user,r=fyko+d2lbbFgONRv9qkxdawL
   S: r=fyko+d2lbbFgONRv9qkxdawL3rfcNHYJY1ZVvWVs7j,s=QSXCR+Q6sek8bf92,
      i=4096
   C: c=<CHANNEL_BINDING>,r=fyko+d2lbbFgONRv9qkxdawL3rfcNHYJY1ZVvWVs7j,
      p=v0X8v3Bz2T0CJGbJQyF0X+HI4Ts=
   S: v=rmF9pqV8S7suAoZWja4dJRkFsKQ=
```

where the ClientProof and ServerSignature have not been recomputed, and where `<CHANNEL_BINDING>` is

```
"c=" + Base64("p=tls-unique,," + CBindData)
```

where `CBindData` is the raw channel binding data. For `tls-server-end-point` it is a hash of the TLS certificate. I don't know yet what the channel binding data is supposed to be for `tls-unique` - it may not even be hashed. I will dig deeper into how `tls-unique` channel binding works.

oberstet:
The 'tls-unique' channel binding is defined as:

> Description: The first TLS Finished message sent (note: the Finished
> struct, not the TLS record layer message containing it) in the most
> recent TLS handshake of the TLS connection being bound to

Eg here is Autobahn https://github.com/crossbario/autobahn-python/blob/master/autobahn/twisted/util.py#L94

ecorm:
I've made changes based on the feedback received for nonces and channel binding. See the commit diff above.

Before merging, I'd like to resubmit as a new PR because I forgot to set my Github email in my local repo.

meejah:
@ecorm if you want to keep the same PR you can always edit the authorship and force-push it (which might make sense 'because lots of comments' -- or maybe that's a reason for a new one :) )

rfc/text/advanced/ap_authentication_scram.md
oberstet: As I am finally going over the PR, wondering, if `id==0x13` corresponds to `version==1.9`, then shouldn't the description `Argon2id variant of Argon2, version 1.3` read => `version 1.9`?

ecorm: From the [latest Argon2 IETF Draft](https://datatracker.ietf.org/doc/draft-irtf-cfrg-argon2):

> This document corresponds to version 1.3 of the Argon2 hash function.

>Version number v is one byte 0x13

If you print 0x13 as decimal, then you'll see 19.

The Argon2 authors decided to use hexadecimal representation for their versioning scheme. Perhaps they wanted it this way so that the second nibble changes only if the major version number changes.

ecorm:
@meejah Thanks for the tip. I'll try it so that I don't lose the commit history.

rfc/text/advanced/ap_authentication_scram.md
oberstet: so actually, the version byte is 2 digit fixed point BCD. wtf? anyways, whatever;)

ecorm:
I was wondering if we should adopt the recommendation of exclusively supporting `tls-unique`, as mentioned in [RFC5929 section 6](https://tools.ietf.org/html/rfc5929#section-6):

> However, it is RECOMMENDED that application protocols making use of TLS channel bindings, use 'tls-unique' exclusively, except, perhaps, where server-side proxies are common in deployments of an application protocol.

I'm not familiar enough with large-scale server deployments to judge if the `tls-server-end-point` channel binding would never/rarely be useful for WAMP.

If we don't know enough right now to make the decision, we can leave both `tls-unique` and `tls-server-end-point` and raise another issue so that we can proceed with the merge.

oberstet:
@ecorm rgd `tls-server-end-point`: at least Autobahn(Python) likely will not have it, but only `tls-unique`, because that will be usable with both WAMP-SCRAM and WAMP-Cryptosign.

ecorm:
@oberstet I repeated the same `tls-unique` recommendation in the WAMP-SCRAM spec, with a link to RFC5929 section 6. That should encourage implementors to use `tls-unique` exclusively, unless they have a good reason to use `tls-server-end-point`.

rfc/text/advanced/ap_authentication_scram.md
ecorm: Ah yes, BDC. That's the term I was looking for. The other alternatives would have been floating point and base-10 fixed point. With base-10 fixed point, you need to divide by 10 to obtain the major version number. With BCD, you only need to do a right shift. The latter might explain their motivation.

ecorm:
I apparently messed-up changing the authorship of my commits. I'm going to dump this PR and will create a new one. Sorry about the mess.

ecorm:
I was able to change authorship of my commits to my Github username. I have no other changes planned for this PR. It is ready to merge unless there are further changes requested by others.

oberstet:
@ecorm you should have permissions to merge yourself, but if you want me to do it, squash&merge, merge commit or rebase to master?

ecorm:
@oberstet I did a rebase and merge because of the early experimental prototypes that were based on my initial draft. This way, the experimenters can see what changed since the initial draft.

ecorm:
Link to WAMP-SCRAM spec here for convenience: https://github.com/wamp-proto/wamp-proto/blob/master/rfc/text/advanced/ap_authentication_scram.md

PR #303, opened 2018-03-15
update generated docs to current master

oberstet:


oberstet:
all generated content .. so merging right away

Issue #182, opened 2015-09-03
Pattern based registrations and multiple matching registrations


oberstet:
With pattern based registrations, define the behavior when a single call matches more than one of a _Callee's_ registrations or more than one registration in general.

konsultaner:
My first thought was that it should only be allowed if [shared_registration] (https://github.com/tavendo/WAMP/blob/master/spec/advanced/shared-registration.md) is supported. Otherwize the registration should have already send an error. 

But thinking of a message monitoring scenario. I would then register to the very base of my uri-set.

    com.project.procedures.shop.buy-withPayPal
    com.project.procedures.shop.buy-withSomethingElse
    com.project.procedures.shop.deleteFromCart
    com.project.procedures.shop.addToCart
    ...
    com.project.procedures.forum.addComment
    com.project.procedures.forum.somethingElse
    ...

If wanted to track all my shop activities, I would additionally register a

    [
        64,
        612352435,
        {
            "match": "prefix"
        },
        "com.project.procedures.shop"
    ]

Or if I want to track all buy calls, I could register

    [
        64,
        612352435,
        {
            "match": "prefix"
        },
        "com.project.procedures.shop.buy"
    ]

---------------

I could also think of a second way to handle this. Thinking in object-orientation the pattern based registration could be something like a callable abstract. So the router could forward all messages to the pattern based registration until there is a concrete or closer match. If this happens the router would send it to the callee that had the better match. 

---------------

I'm not sure what to prefer but I cannot think of a scenario where the second attempt would make sense. So I would say: send the call to all matches.

konsultaner:
Is this a problem of WAMP or a business logic problem? Imagine you provide the pattern based registration and shared registration to your callees. One Callee registers a concrete procedure like ```com.example.procedure.something``` and also adds a procedure to track all procedure calls with ```com.example.procedure..```. If another callee only adds ```com.example.procedure.something``` and not ```com.example.procedure..```, the wildcard procedure would be call only every second call if the policy was round robin.

I'd say this is a business logic problem...

mbonneau:
For all use cases I can come up with (which I am sure is incomplete), it seems that URI parts are basically becoming arguments to the RPC.

The complexity added by having to figure out how to handle this issue where both the ```CALL``` and the ```REGISTER``` are able to use pattern/prefix-based URIs is too much to make it worth it.

The best solution is not allowing this on ```REGISTER```. Just move the "argument" part of the URI to the arguments. A call:
[...]

PR #305, opened 2018-03-27
Described pattern-based RPC call routing

KSDaemon:
Tried to combine all our thoughts and extend it with examples.
Refs #182


oberstet:
Are we certain, that this set of rules always leads to either exactly one best matching registration or "procedure not found"?

KSDaemon:
I do think so! :) 
@oberstet do you think, it's not clear right now? Should define more clearly?

oberstet:
LGTM! So, not sure if we agreed upon "required upvotes", but in any case, the material in the PR is a summary of what was discussed broadly before. So I am merging this;)

Issue #314, opened 2018-06-19
Session meta procedure to forcibly disconnect a session


gammazero:
Since there have been requests for such a feature I wanted to propose this as a part of the WAMP specification.  This is already supported by Crossbar: https://crossbar.io/docs/Session-Metaevents-and-Procedures/#killing-a-session, and seems like something that should have been proposed.  I could not find mention of such a feature in any issue, so I am opening an issue.

The proposal is for a session meta procedure, `wamp.session.kill`.  It takes one positional argument, `session|ID`, two optional keyword arguments, `reason` and `message`.  The `reason` argument is the URI used as the `GOODBYE.Reason`. The `message` argument is put into `Goodbye.Details` under the key "message".  If the session does not exist, an error is returned.

ecorm:
There should also be a `wamp.session.kill_all` meta procedure for closing _all_ of the sessions associated with an `authid`, as I proposed in https://github.com/crossbario/crossbar/issues/1340. Two use cases in our application motivated the need for this:

- The application admin removing a user account at runtime.
- The application admin suspending (deactivating) a user account at runtime.

We are currently resorting to our application backend keeping track of all frontend sessions and associating each of the active sessions with the `authid`. Our backend must then call Crossbar's `wamp.session.kill` meta procedure in a loop for each active session associated with the `authid`.

gammazero:
I am currently planning on implementing both of these as stated in https://github.com/gammazero/nexus/issues/112.  So, it sounds like these are both reasonable extensions of the WAMP session meta api.  If there is consensus on this, I will draft the spec documentation of these.

oberstet:
thinking about, since the main WAMP session related information is:

* `session_id`
* `authid`
* `authrole`
* `realm`

we could go for 4 procedures:

* `wamp.session.kill`: kill a single specific session given a WAMP session ID
* `wamp.session.kill_by_authid`: kill all currently connected sessions with given WAMP authid
* `wamp.session.kill_by_authrole`: kill all currently connected sessions with given WAMP authrole
* `wamp.session.kill_by_realm`: kill all currently connected sessions for the given WAMP realm

gammazero:
I have given it some thought, and this seems like a solid plan.  It is simple, and is addresses all the identified use cases.   It has my vote!

ecorm:
[...]

PR #315, opened 2018-06-22
Describe wamp.session.kill meta procedures

gammazero:
This change documents four new meta procedures for the session meta API:

- wamp.session.kill
- wamp.session.kill_by_authid
- wamp.session.kill_by_authrole
- wamp.session.kill_all

Fixes issue #314

### DIFF
--- a/rfc/text/advanced/ap_session_meta_api.md
+++ b/rfc/text/advanced/ap_session_meta_api.md
@@ -39,11 +39,15 @@ Fired when a session leaves a realm on the router or is disconnected. The event 
 
 #### Session Meta Procedures
 
-A client can actively retrieve information about sessions via the following meta-procedures:
+A client can actively retrieve information about sessions, or forcefully disconnect sessions, via the following meta-procedures:
 
 * `wamp.session.count`: Obtains the number of sessions currently attached to the realm.
 * `wamp.session.list`: Retrieves a list of the session IDs for all sessions currently attached to the realm.
 * `wamp.session.get`: Retrieves information on a specific session.
+* `wamp.session.kill`: Kill a single session identified by session ID.
+* `wamp.session.kill_by_authid`: Kill all currently connected sessions that have the specified authid.
+* `wamp.session.kill_by_authrole`: Kill all currently connected sessions that have the specified authrole.
+* `wamp.session.kill_all`: Kill all currently connected sessions in the caller's realm.
 
 Session meta procedures MUST be registered by the *Router* on the same realm as the WAMP session about which information is retrieved.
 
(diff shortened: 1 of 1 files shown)


gammazero:
Please consider whether the `wamp.session.kill...` procedures are best suited as additional procedures in the `session_meta_api` or if they are better suited as a separately announced feature, such as `admin_meta_api`.

rfc/text/advanced/ap_session_meta_api.md
```diff
@@ -39,11 +39,15 @@ Fired when a session leaves a realm on the router or is disconnected. The event
 
 #### Session Meta Procedures
 
-A client can actively retrieve information about sessions via the following meta-procedures:
+A client can actively retrieve information about sessions, or forcefully disconnect sessions, via the following meta-procedures:
```
ecorm: Should say "forcefully **close** sessions", instead of disconnect. The spec uses the terms "connect"/"disconnect" within the context of transports.

ecorm:
Should the transport associated with the killed sessions be disconnected as well? Or should only the session be closed?

gammazero:
@ecorm The use cases that I see for this would want the transport disconnected.  For example, I change the certificates that my router trusts and I want to kick all the remote clients off to force them to reconnect over TLS connection.  There may be legitimate use cases for only closing the session and leaving the transport, but I think those are too much edge cases to merit that degree of control, and would generally still be ok if they had to reconnect the transport.

Similarly, I want to avoid having kill exclusion lists, or kill black/white listing, etc.

Oh, I will also add the following detail to the `wamp.session.kill_all`:
"Closing all sessions in the realm will not generate session meta events, since no subscribers would remain to receive these events."

gammazero:
@ecorm If this needs more discussion before merging any documentation, I can close PR for now until there is greater consensus and discussion of detail.

oberstet:
the meta API is exclusively concerned with WAMP session - not transports. It is an implementation detail left unspecified whether a transport is closed when a session leaves or not.

IMO, in the spec, we should also try to use precise language: there is not such thing as a "connected" session. A session is _joined_ on a realm or not. it can proactively decide to _leave_ the realm - or be kicked ..

ecorm:
According to the [statechart](https://wamp-proto.org/static/rfc/draft-oberstet-hybi-crossbar-wamp.html#rfc.section.7.2.2), the resulting states would be _closed_ / _established_ .

_Leave_ is a verb that the subject performs upon oneself. You can't say "I leaved him from the bar", when you mean to say "I kicked him from the bar", or "I made him leave the bar".

Client Action | Resulting Session State
----- | -----
join | established
leave | closed

Router Action | Resulting Session State
----- | -----
kill | closed

In hindsight, "kick" might have been a better verb than "kill", especially when you think about chatrooms or multiplayer games. "kill" is rather something to do to an operation system process. However, I acknowledge that there are already router implementations out there that use the term "kill" in the meta-procedure that forcefully closes a session.

gammazero:
If no `reason` is given, should this default to `wamp.error.system_shutdown` or should a GOODBYE message be sent without a reason?

oberstet:
@gammazero default reason should be `wamp.close.normal`. sending GOODBYE without reason is a protocol error.

oberstet:
LGTM .. Emile also reviewed it .. so I am gonna merge this. Should there be sth missing, we can come back, but for me, this looks pretty precise and complete. cool!

Issue #316, opened 2018-07-06
Testament API scope keyword `detatched` is misspelled


gammazero:
The WAMP specification specifies one of the Testament scope keywords as `detatched`.  This should be spelled `detached`.

It appears that Crossbar uses the spelling `detatched`.  The [nexus](https://github.com/gammazero/nexus) router uses the spelling `detached`.  I do not know if any other WAMP routers implement the testament API.

oberstet:
@gammazero rgd crossbar: yeah, typo .. https://github.com/crossbario/crossbar/issues/1362

PR #318, opened 2018-07-06
Fix spelling of testament scope keyword "detached"

gammazero:
The WAMP spec spells the testament API scope keyword as `detatched`.  The change corrects the spelling to `detached`.  Since the original spelling is already in use in at least one router implementation, there may be a need to provide some note about this change.

Fixes issue #316


oberstet:
no idea how this could actually went unfixed up to the implementation in crossbar! anyways: crossbar has a bug now .. and will fix it. https://github.com/crossbario/crossbar/issues/1362

Issue #287, opened 2017-09-07
More information for wamp.subscription.on_delete


i2xS:
same problem as here https://github.com/wamp-proto/wamp-proto/issues/238
Currently `wamp.subscription.on_delete` only receives `subscription_id` as argument and there is no chance to get uri of this subscription because calling `wamp.subscription.get` in `on_delete` handler with this id raises `wamp.error.no_such_subscription`

oberstet:
No, this isn't the same. A subscription isn't identified by a URI.

PR #320, opened 2018-07-10
Add arguments for wamp.session.on_leave and wamp.subscription.on_delete

### LINKED ISSUE #238: More informations for wamp.session.on_leave
Currently `wamp.session.on_leave` only receives the session id as sole argument. This imposes a problem, as a subscribee has no way to acquire additional informations he might be interested in like, authid, role or the realm left.

But wait, isn't there `wamp.session.get` you might ask now. Indeed there is, but the current draft makes no assertions about whether the session information can still be queried in `wamp.session.on_leave` and the standard implementation crossbar will currently let you down in that regard and correctly tells you, that he doesn't have such a session, as it has already been left.

gammazero:
This PR fixes #238 and #287 by providing additional information on items that are removed from the realm, making that information otherwise unavailable.

### DIFF
--- a/rfc/text/advanced/ap_pubsub_subscription_meta_api.md
+++ b/rfc/text/advanced/ap_pubsub_subscription_meta_api.md
@@ -70,11 +70,11 @@ Fired when a session is removed from a subscription. The event payload consists 
 
 ###### wamp.subscription.on_delete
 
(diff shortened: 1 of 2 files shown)


oberstet:
I am fine with the change rgd #238, but not #287

PR #328, opened 2018-10-05
Fix typos and clarify some grammar

LaurenceGA:
Just some clarifying changes.
- The description for section 9 (Publish and Subscribe) says that it defines implementation details for a *dealer* where it should say *broker*. It even says it's for a *broker* in the next sub-section.
- A couple of grammar and sentence rearrangements


KSDaemon:
Hi @LaurenceGA Thanks for fixes! 
Seems to be ok!

PR #330, opened 2018-10-19
Add publisher_disclose on subscriber side

martin31821:
Currently, we have a disclose_caller on the callee side, but there is no equivalent on the susbcriber side.


oberstet:
This has been discussed many times before, and rejected. If you want to discuss it, bring up reasons etc, please file an issue first .. then we can also link all the old discussions there rgd the topic

PR #347, opened 2019-03-08
Update D-Bus information

FGasper:


oberstet:
not sure I would call what Dbus provides "routed RPC" .. or "routed IPC" for that matter. doesn't Dbus require the bus address (a network level thing) of the callee to be provided when doing a call?

FGasper:
> doesn't Dbus require the bus address (a network level thing) of the callee to be provided when doing a call?

It requires a bus name, which can be either the unique name (assigned at connection) or, more commonly, a service name. The service name is the “routed” part: it could be on the same machine or across a network.

The workflow for requesting/registering a D-Bus service name is the same as how WAMP registers RPC methods.

oberstet:
Ok, I see! In that case, sure, according to our own table, there should be a check on the "routed RPC" column as well. I think it could be worth explaining Dbus mechanisms and the correspondance to WAMP, etc a bit more even on a dedicated section on the web page (or the FAQ).

and also, the obvious follow up Q: why do we need WAMP when there is Dbus? ;) or: can we define a seamless bridging between? etc

oberstet:
Somehow this got lost on me too: while I was aware of `p5-Net-WAMP` of course, `p5-Protocol-DBus` is new to me - cool! I guess given you have implemented both WAMP and DBus, you'll be in a perfect position to draw comparisons/conclusions - that would be super interesting!

FGasper:
Yeah I learned D-Bus more recently.

D-Bus serves a lot of the same use cases as WAMP. The bonus is that on a lot of systems—pretty much all modern Linuxes—it already runs by default. It’s optimized for applications that run entirely on the same server, though: you have to distinguish signed from unsigned, etc.

Its maintainers have mostly disavowed the potential for use across a network, though. The logic is still there, but for example, there’s no authentication mechanism. Its custom serialization is also nontrivial, which makes it harder to implement a simple D-Bus endpoint than WAMP— of course, most boxes that run D-Bus also expose at least one C implementation, so applications can pull that in.

Does that clarify?

-F

oberstet:
yeah, thanks a lot for the explanations and perspectives!

PR #357, opened 2019-08-16
add WampSyncClient to implementations

jszczypk:
The only current implementation of WAMP for PHP is Thruway made using asynchronous React event loop, which is for example hard to use in classic-synchronous PHP code.
Here is a small library with minimal dependencies that allow for two line call to remote WAMP procedures as well as publish WAMP events.


oberstet:
Nice! Thanks for sharing your work=) The docs clearly state only caller and publisher roles are supported - perfect.

PR #359, opened 2019-08-17
Fix number of messages in bp_messages.md

codebrainz:
I thought maybe the 25 was also including the Advanced messages, but there are only 4 of those and the surrounding text here clearly refers to the Basic messages. Or did I miss something?


KSDaemon:
Maybe we can just omit the number of messages? I don't think it is a very important number, the word “all” is enough. What do you think?

Issue #352, opened 2019-06-19
callee wamp.error.unavailable + transparent call rerouting


binaek89:
This is to enable the ability of a RPC service provider to `REJECT` an `INVOCATION` request coming in.

This message is transmitted solely from the `CALLEE` to the `DEALER`.

When the `DEALER` receives this message from the `CALLEE`, it will select another `CALLEE` to process the original `CALL` request which corresponds to the original `INVOCATION` request. 

The original routing rules still apply. If the `DEALER` runs out of `CALLEE`s or is unable to select a new `CALLEE` (when `first` or `last` is the routing logic selected), it returns an `ERROR` message to the `CALLER`.

### Use Case
A use-case is when a `CALLEE` is too heavily loaded to be able to process a request or when the application instance running the `CALLEE` service is aware that it will be shutdown as a part of a scale-in exercise by the orchestrator - wherein, the instance is expected to finish execution of running executions, but should not take up new jobs. 

This will enable the `DEALER` to handle such a case without the `CALLER` ever having to know about it.

### Suggested change in specification

This will require a new `Message Code`.

An extension to the protocol could take the form:

#### Invocation REJECT

If the Callee purposefully wants to reject execution of the invocation, the Callee sends a `REJECT` message to the Dealer:
```
[REJECT, INVOCATION, INVOCATION.Request|id, Details|dict]
```

where

* `INVOCATION.Request` is the ID from the original `INVOCATION` request previously sent by the Dealer to the Callee.
* `Details` is a dictionary with additional details.

*Example*
Assuming `Message code` to be `71`
```
[71, 68, 6131533, {}]
```

mbonneau:
Having `REJECT` would also allow callees to fend off new requests if it is shutting down. This would be transparent to the caller.

May help with the use case I describe here:  https://github.com/wamp-proto/wamp-proto/issues/283

This would allow the callee session to finish up all `INVOCATION`s and then `UNREGISTER`. Although in that case it is still possible that the `UNREGISTER` is sent with `INVOCATION`s in transit.

binaek89:
> Having `REJECT` would also allow callees to fend off new requests if it is shutting down. This would be transparent to the caller.
>
That is the use case I was suggesting.
> 
> May help with the use case I describe here: #283
> 
> This would allow the callee session to finish up all `INVOCATION`s and then `UNREGISTER`. Although in that case it is still possible that the `UNREGISTER` is sent with `INVOCATION`s in transit.

[...]

PR #365, opened 2019-11-22
Transparent call re-routing

binaek89:
This is the draft sprecification for issue #352 .

In addition to the `wamp.error.unavailable` URI, I added another `wamp.error.no_available_callee` which is the line of communication between the *Dealer*/*Router* to the *Caller* in case all available registrations respond with `wamp.error.unavailable`.


binaek89:
@oberstet did anyone get a chance to look at the changes?

oberstet:
some quick comments:

* IMO, this feature should be described as part of / additional section to the existing text for shared registrations: https://wamp-proto.org/_static/gen/wamp_latest_ietf.html#shared-registration 
* support (in principle) for this AP feature is optional, hence we need to spec feature announcement (https://wamp-proto.org/_static/gen/wamp_latest_ietf.html#router-role-and-feature-announcement)
* further,  using the feature could require option(s) given by the callee when registering, eg `REGISTER.Options.max_reroutes|int`

eg `max_reroutes == 0` would be the default leading to the existing behavior, while `max_reroutes > 0` would allow to retry on a different callee (than the one that returned `wamp.error.unavailable`) up to the specified maximum number of retries

gammazero:
This looks good.  There is one other failure scenario that maybe the spec should call out how to handle:  When the first _Callee_ returns `wamp.error.unavailable`, and a subsequent _Callee_ returns some other error or fails to respond.  Should the call be re-routed to the next _Callee_, or should the invocation be terminated with the error from the failed _Callee_?

binaek89:
> This looks good. There is one other failure scenario that maybe the spec should be more explicit about: When the first _Callee_ returns `wamp.error.unavailable`, and a subsequent _Callee_ returns some other error or fails to respond. Should the call be re-routed to the next _Callee_, or should the invocation be terminated with the error from the failed _Callee_?
> 
> I think the latter is implied by the text,
> 
> > In case all available registrations of a _Procedure_ responds with a `wamp.error.unavailable` ...
> 
> but asking if anyone else thinks that needs clarification.

The way I see it, any error other than `wamp.error.unavailable` should terminate the invocation (almost as if the re-routing feature does not exist at all)

oberstet:
> The way I see it, any error other than wamp.error.unavailable should terminate the invocation (almost as if the re-routing feature does not exist at all)

yeah, agreed. that would have also been my expectation. the `wamp.error.unavailable` should be reserved as a specific error indicating that no callee could be reached at all while the procedure does exist, so it's different from "no procedure registered, which would only happen when all callees have unregistered)

oberstet:
so I'm +1 on merging, which makes it +3 votes in total. I merge this! should there be more, let's do that on a subsequent PR ..

PR #367, opened 2020-03-09
Add WAMPex

entone:
Adds WAMPex to the list of client implementations. WAMPex is a pure Elixir implementation of a WAMP client.

### DIFF
--- a/docs/implementations.rst
+++ b/docs/implementations.rst
@@ -57,6 +57,7 @@ Swamp_              Swift                               WAMP client library in p
 wampcc_             C++                                 A C++ WAMP library that aims to depend only on C libraries. Includes a basic router.
 WAMP_POCO_          C++                                 A fork of AutobahnCpp that is using `POCO <http://pocoproject.org/>`_ instead of Boost.
 wamped_             C++                                 Intended for running on `ARM mbed OS <https://www.mbed.com/en/>`_. Experimental.
+WAMPex_             Elixir                              Elixir client library that implements the complete client Basic Profile and much of the Advanced Profile. Defaults to msgpack over WebSocket, but JSON is supported as well as raw TCP sockets.
 WampSyncClient_      PHP                                 Synchronous (blocking) PHP client implementing Caller and Publisher roles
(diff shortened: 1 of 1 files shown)


oberstet:
cool! yet another implementation, nice:)

I tried looking a bit at https://gitlab.com/entropealabs/wampex/-/blob/master/priv/client.json .. this is the core of the state machine thing, right? the intro text mentioned it, and it caught my eye .. wondering: is that supposed to be complete rgd initial session establishment .. incl authentication?

> PS: the CI has a fail that is unrelated .. gonna merge this one and fix the issue separately

entone:
It's the "base" of the state machine implementation, other events trigger the state transitions. 

image

it's nice to give a visual representation to the states, and has a nice interface for actually implementing the state machine. There's still quite a bit of code to build out the actual implementation, https://gitlab.com/entropealabs/wampex/-/blob/master/lib/client/session.ex

But the StatesLanguage lib manages all state transitions and ensures you can't get into a bad state. It's also completely asynchronous.

I'm building up a similar library for a router as well.

oberstet:
^ thx for the charts and explanation! looks cool. wondering: it seems the boxes (states?) represent (mostly) WAMP messages while the arrows in between represent messages (received)?

also, not sure you saw this, we have a statechart for session establishment and authentication (but only that .. eg not for all of wamp in general):

https://wamp-proto.org/_static/gen/wamp_latest_ietf.html#rfc.section.7.2.2

entone:
I have seen that.

If you think about the state machine as a graph, specifically a DAG, the
arrows are your edges, and the boxes are your nodes. The edges are some
sort of event that has occurred, in this case, more than likely a message
has arrived over the transport, and causes a state transition aka move to
another box. Once a state has done it's work, usually adding some data to
the internal state of the process, it fires another event(arrow) and we
move to the next state.

Internally we're utilizing an Erlang OTP process called gen_statem, you can
read more about it here, http://erlang.org/doc/design_principles/statem.html

StatesLanguage is a macro that parses the StatesLanguage JSON format, and
creates a gen_statem module that provides callbacks for different
gen_statem callbacks.

I'm the creator of that library as well.

There's some pretty robust docs for the StatesLanguage library at
https://hexdocs.pm/states_language/readme.html

PR #368, opened 2020-03-22
updates the title, description and link to the WAMPexClient library

entone:
Moving the repositories around a bit as I make room for the router implementation


oberstet:
thanks for the update!

merged, and auto-deploy worked https://wamp-proto.org/implementations.html#libraries

entone:
Awesome! Will probably have a PR for the Router library later in the week.

PR #374, opened 2020-09-06
Added a new section to track useful utilities like CLI tools to interact with WAMP implementations

frol:


oberstet:
awesome: thanks for contributing! yeah, I like the idea of having a separate section for "utilities", which definitely is sth different than "client libraries" and "routers".

PR #375, opened 2020-09-09
Add asynchronous WAMP client library implementation in Rust

frol:
This implementation is not feature-complete, but it is in a usable state for pet-projects. I am going to contribute to the project while building my pet-projects using it.


KSDaemon:
Hi @frol Thanks for contributing!

PR #378, opened 2020-10-20
Editorial Correction - "an URI" for "a URI"

duckboy81:
Pretty simple, just corrected the grammar for a few of the doc pages. Swapped any reference of `an URI` for `a URI`.

Thanks to those of you working on this! My addition is peanuts in comparison.


oberstet:
welcome, and thanks for contributing! 

> "pretty simple ..": well, not for non-native speakers;) seriously, why is it that the usual "an vs a" rule is not followed in this case? indeed it seems to be more complex than I thought;) fwiw, found this explanation https://lang-8.com/uj/journals/151322

PR #381, opened 2021-02-03
Remove excess HTML tag <td> from documentation

CanNuhlar:
On  Routers section of Implementations page, there was an excess tag of <td> on wamprt table item. This commit removes that tag and corrects the visibility.


CanNuhlar:
That was fast!

KSDaemon:
I'm a little bit late :)

PR #384, opened 2021-10-21
WAMP DAO proposal

oberstet:
Please see https://github.com/wamp-proto/wamp-proto/issues/387


oberstet:
I'm squashing and merging what we have on this PR now, even though the documents added in the PR are not fully complete, agreed/discussed. np, we can change as needed

the PR doesn't touch any of the preexisting content, only adds new folders "dao" and "wep". so I think merging should be fine and non-controversial. as said, rgd the new DAO related stuff: this is still to be seen "under discussion" ..

PR #388, opened 2021-12-02
A few updated to w3c respec build: fixed warnings

KSDaemon:
A small update to w3c spec builder: fixed errors and warnings. Hope it will works as expected :)


oberstet:
cool, yeah, it should work;) not sure if all of the CI does run smoothly currently, but if not, that isn't related to this PR of course. I'll be on the call today half an hour earlier (1830 CET) in case you or anyone wants to catch up with the 1st call ..

PR #403, opened 2022-06-17
Deploy refactoring

KSDaemon:
Hi folks!

Whats on plate: 

- Refactored all the mess in spec build
- Fixed building spec in w3c format
- Now editors of the spec are gathered during build from github log records :)
- fixed some typos and code block
- Set up deploy to gh-pages

What do you think? Please, have take a minute to have a look


KSDaemon:
Right now action is set up to run on every push. But if everything is ok — i'll correct it to run only on push to master. So we can have rather stable version publicly available

oberstet (approved):

oberstet:
that is awesome! you are a hero=) moving to a (again) working auto-build for the spec text and web site, and auto-deploy - while letting me retire the AWS stuff (because it uses GH pages) is fantastic! anyways, I'm voting +1

oberstet:
as this isn't changing anything rgd the spec substance matter, but only affects the technical build and hosting approach, and since there hasn't been anyone objecting, I'm merging this now ...

oberstet:
follow up for the dns stuff https://github.com/wamp-proto/wamp-proto/issues/404

oberstet:
sidenote: can we delete https://github.com/wamp-proto/wamp-proto/tree/deploy-refactoring now it is merged? after that, I'd like to disable new branches other than `master` or `gh-pages` if possible (thus kinda enforcing the standard GH PR workflow with `master` the only target branch ..)

KSDaemon:
@oberstet yeah, i personally prefer to delete source branch while merging :) 
So we can delete this one of course. I'll make a new if needed :)

You mean you want to disable direct push to protected branches? — If so, agree it is a good idea. I think we can not do this for `gh-pages` branch, as during push to master our gh action is preparing and pushing updated content to that branch. Need to figure out how it can be protected.

oberstet:
> So we can delete this one of course. I'll make a new if needed :)

done

> You mean you want to disable direct push to protected branches?

yes, exactly. I guess we're thinking along similar lines .. yes, _ideally_ `gh-pages` would also be protected. but I get the problem. in any case, yeah, enforcing the GH PR workflow with mandatory green CI (all GH actions are running good) would be great IMO. it is self-imposed discipline that is enforced without doing manual work .. good=)

KSDaemon:
> it is self-imposed discipline that is enforced without doing manual work .. good=)

Totally agree! :)

PR #405, opened 2022-06-21
WEP007 - WAMP API Schema Definition with Flatbuffers

oberstet:
A proposal for a WAMP IDL, plus schemas for the following WAMP APIs, part of which are standard:

* [x] WAMP IDL
* [x] WAMP Meta API
* [ ] WAMP Dynamic Authenticators and Authorizers API

> not part of this PR are: Dynamic Node Control API and E2E Encryption Key Exchange API (under development).

---

A demo of the features described here, using Crossbar.io/Autobahn as a platform, can be found here https://github.com/crossbario/crossbar-examples/tree/master/payload-validation


oberstet:
one thing that I would like to do is to refer to the WAMP meta API schema definition from the respective spec text, eg like:

grafik

now, I'm not sure what the best way would be

* link-and-include-contents
* link-bundled-schemafile
* link-to-referenced-item
* ...

KSDaemon:
> cleaning up the text/sections from the advanced profile, I now think we should split up the BP and AP into 2 documents:

@oberstet well, why not? But maybe we can put all options on this page: full spec + bp + ap? So users can choose what they want? It's just a matter of the build process.
I personally prefer to look through the whole spec in one doc. For new users it may be easier to dig into BP first.

oberstet:
@KSDaemon that's a good idea! yeah:

* full (BP+AP) spec
* BP only
* AP only

oberstet:
I've added PDF as output format in addition to HTML and TXT. requires the stuff from `make requirements_xml2rfc`, so this needs to be avail on GH actions as well. there is BP-only, AP-only and FULL flavors
Bildschirmfoto von 2022-06-26 22-13-09

Bildschirmfoto von 2022-06-26 21-14-56

Bildschirmfoto von 2022-06-26 22-10-21

oberstet:
> I just don't see html w3c format.

yeah, it is still in the Makefile but I removed it from the web site page (for now), since I don't want to always check 2 target generators (the generated contents), and because this spec is IETF anyways ..

KSDaemon:
Hmm.... I check IETF html version and it looks pretty good (as compared to one a few years ago). So I agree to remove w3c format. I'll remove it and cleanup makefile and repo. Just will wait until your PR will be merged just to pass conflicts :)

oberstet:
@KSDaemon sounds good, awesome!!

> Just will wait until your PR will be merged just to pass conflicts :)

Ok, I've just pushed all my changes, and it looks quite decent IMO

not sure how we should proceed rgd consensus among implementors. On the other hand, the changes within this PR contain:

1. lots of polish without changes to substance matter
2. a strictly new chapter "WAMP Interfaces" 

the first part shouldn't require consensus .. it is just making the spec nicer to consume

the second part is WIP and experimental, even though crossbar/autobahn now have implementations. rgd that part, we could for now technically add (merge) the new chapter, then discuss and gather feedback, and in case we disagree, remove the whole chapter again ...

thing is: as you hint, would be good to get my polish stuff merged so that it doesn't slow down eg you or others working on the spec or build scripts

me personally, I want to work on the yellow marked stuff (on a new PR once this is / would be merged)
Bildschirmfoto von 2022-06-27 21-32-00

Makefile
```diff
@@ -70,38 +70,69 @@ start_build:
...
-build: clean start_build build_images build_spec sphinx-build-docs
+build: clean start_build build_images build_spec build_docs
 
 build_images: $(SITEBUILDDIR)/$(OBJECTS)
 
 build_spec: build_spec_rfc build_spec_w3c
 
+# PS and RTF support are provided using a combination of wkhtmltopdf, pdf2ps and/or GNU enscript.
+requirements_xml2rfc:
+	sudo apt install xml2rfc enscript
```
KSDaemon, line 81: I think it's not a good idea to mess installing OS packages in runtime.
First point — it breaks down using this makefile on different OS. For example mac os users like me can not simply run this makefile.
And second point: it's better to put OS requirements in gh-action steps. So it can be effectively cached without run every time.

Makefile
```diff
@@ -70,38 +70,69 @@ start_build:
...
+	pip install 'pycairo>=1.18' 'weasyprint<=0.42.3'
+
 # https://github.com/mmarkdown/mmark
 # sudo apt install -y mmark
 requirements_mmark:
 	wget https://github.com/mmarkdown/mmark/releases/download/v2.2.25/mmark_2.2.25_linux_amd64.tgz
```
KSDaemon, line 87: Right now it's workable only on linux. My tech debt. Also thinking to make it cross platform

Makefile
```diff
@@ -70,38 +70,69 @@ start_build:
...
 	tar xvzf mmark_2.2.25_linux_amd64.tgz
 	rm -f ./mmark*.tgz
+	sudo cp ./mmark /usr/local/bin/
```
KSDaemon, line 90: Any special reason to run it with sudo? I think we can also move this to gh-action steps. So any user can just install it on own OS. So we need just to check for it.

KSDaemon (commented):
Big work! 
Just have some comments regarding makefile.

Makefile
oberstet: I've tried to improve the PR rgd your comments (pls see last commit). having said that, you can ignore those makefile targets (these are _not_ run as part of actual build targets), they are mainly to help my memory / for convenience (when on linux) ..

Makefile
oberstet: ok, done. not sure if it works ... we'll see

KSDaemon:
👍🏻  We'll see :)

KSDaemon:
@oberstet we prefer to squash commits?

oberstet:
> we prefer to squash commits?

yes, in my view. this leads to a clean history. so actually, I'd opt for "squash & rebase". the rebase is to put the changes at the end of history of current master. in case of the PR here, I think there aren't any changes on master after I forked for this PR ...

> We'll see :)

the current GH actions only run on the push after merge? we could think of running (that is, building without publishing to GH pages) the build already while the PR is open, and then disallow merging unless CI is green ...

oberstet:
fantastic! thanks=)

> btw, rgd "preserving change history" and squash&rebase: yes, that is the main argument against. now, at least for how I personally carve out new stuff and work with commits on a PR, it is often very chaotic;) I use "commit & push" on a WIP PR mainly as kinda glorified "save all" button. It gives me peace in mind .. should my house explode accidently, the code is safe!! even though I might be dead ... haha ... _why_ does that thinking comfort me I have no clue. anyways. bla bla bla;)

oberstet:
now trying to figure out

```
 mmark ./rfc/wamp.md > ./.build/wamp.xml
/bin/sh: 1: mmark: not found
make: *** [Makefile:92: build_spec_rfc_mmark] Error 127
Error: Process completed with exit code 2.
```

https://github.com/wamp-proto/wamp-proto/runs/7101833203?check_suite_focus=true#step:12:102

KSDaemon:
You occasionally removed one target:
```
# https://mmark.miek.nl/post/syntax/
---build_spec_rfc: requirements_mmark build_spec_rfc_mmark
+++build_spec_rfc: build_spec_rfc_mmark build_spec_bp_rfc_mmark build_spec_ap_rfc_mmark
```

KSDaemon:
You can just put it back. I'll refactor it in mine branch

oberstet:
almost;) just some left over

```
 Created file ./dist/wamp_ap_latest_ietf.pdf
make: *** No rule to make target 'build_spec_w3c', needed by 'build_spec'.  Stop.
Error: Process completed with exit code 2.
```

oberstet:
fixed without PR https://github.com/wamp-proto/wamp-proto/commit/d1e7c52fa8ca0c8e5d27f4fc0f2eada1aa5f9413

PR #406, opened 2022-06-21
Deploy refactoring

KSDaemon:
- Removed deploy to AWS
- Set to rebuild site/spec on on push to master (rather than on every commit)
- Some polishment


KSDaemon:
Ough! Because i've continued to use same branch it shows so much commits. But it is only one :)

PR #407, opened 2022-06-21
Deploy refactoring

KSDaemon:
- Removed deploy to AWS
- Set to rebuild site/spec on on push to master (rather than on every commit)
- Some polishment


KSDaemon:
Here is clean one!

PR #408, opened 2022-06-28
Site polishment

KSDaemon:
Some site polishments and refactoring.
Completely removed w3c spec, grunt and other obsolete things.


KSDaemon:
@oberstet Don't waste time fixing broken gh action. I think it should be fixed with this one.

PR #410, opened 2022-06-28
Deploy fixes

KSDaemon:


oberstet:
awesome!! works, incl pdf

Bildschirmfoto von 2022-06-29 00-43-23

oberstet:
@KSDaemon one Q rgd GH pages: this requires the _built_ site files (incl. html and pdf) to be created on the `gh-pages` branch - and hence dump and add build files to the git repo? if so, then IMO this sucks slightly: could we move GH pages to a different _repo_? if not, why are people fine misusing git as a build dump rather than a source repo?

KSDaemon:
@oberstet Absolutely agree with you. Don't know why GitHub decided to choose that approach. 
Let's think about it. Maybe we could find a solution. If not — we'll move building the site to another repo. Hope we'll manage access to other repo.

PR #411, opened 2022-06-28
Updates to index page

KSDaemon:
Added words about serializers and transports to index page


KSDaemon:
Yeah! Sorry! Just a small fix. Deleted and cleaned up.

PR #412, opened 2022-06-28
add test CI for PRs

oberstet:


oberstet:
ok, so I've added a "Build spec and website" CI step that runs on PRs, and which is required to run green before merging, and also at least 1 review (we can disable that of course .. let's discuss!)
Bildschirmfoto von 2022-06-29 01-55-33

KSDaemon:
Awesome! @oberstet You rock! 
I'm +1 on review and approving. Even if it is a rather clear change — it will be good to have another pair of eyes looked through it.
I think we can bypass this rule in case of small oneline fixes

oberstet:
> I think we can bypass this rule in case of small oneline fixes

yeah, agree (and right now, admins can still by-pass and push-direct etc)

PR #413, opened 2022-06-29
start describing wamp-cryptosign [WIP]

oberstet:


oberstet:
ok, so the basic WAMP-Cryptosign authentication is now described in this PR, ready for review/feedback:

grafik

the description of basic authentication via WAMP-Cryptosign reflects the implementation in different libraries, which are practically used / interoperate as of today.

beyond the basic authentication, the following optional authentication features will be described:

* [ ] TLS Channel Binding
* [ ] Router Authentication **new**
* [ ] Trustroots and Certificates **new**


Router authentication allows the client to authenticate the router, using the same signature mechanism as used by the client to authenticate to the router.

Trustroots and certificates allows the router to trust a client with a new, unknown public key if a certificate chain covering that public key is present, and when the trustchain ends in a trustroot accepted by the router.

rfc/text/advanced/ap_35_authentication_cryptosign.md
```diff
@@ -1,3 +1,937 @@
...
+
+The `HELLO.Details.authextra|dict` contains the following members for WAMP-Cryptosign:
+
+| Field             | Type     | Required | Description                                                                          |
+|-------------------|----------|----------|--------------------------------------------------------------------------------------|
+| pubkey            | string   | yes      | The client public key (32 bytes) as a Hex encoded string, e.g. `545efb0a2192db8d43f118e9bf9aee081466e1ef36c708b96ee6f62dddad9122` |
+| channel_binding   | string   | no       | If TLS channel binding is in use, the TLS channel binding type, e.g. `"tls-unique"`. |
+| challenge         | string   | no       | A client chosen, random challenge (32 bytes) as a Hex encoded string, to be signed by the router. |
+| trustroot         | string   | no       | When the client includes a client certificate (see below), the Ethereum address of the trustroot of the certificate, e.g. `0x72b3486d38E9f49215b487CeAaDF27D6acf22115` |
```
KSDaemon, line 85: I think we should add more words about trustroots. Maybe describe possible cases, like the most common one is to rely on underlying (browsers, OS) root certificates. But this option allows users to choose and specify independent blockchain-based authority for trusted certificates. And maybe also say, that Ethereum is not the one possible option, that this can be extended in the future.
Because i think some users may be afraid that they need blockchain for wamp to work %) Which is of course not true :)

KSDaemon (changes requested):
Leaved some comments to fix/think of.

oberstet:
@KSDaemon awesome, working on it!

-----

(unrelated): probably add text, since we are following a similar approach:

> Host Identity Protocol Version 2 (HIPv2) https://datatracker.ietf.org/doc/html/rfc7401 => Host Identity (HI) and Its Structure

> Briefly, the HIP architecture proposes an alternative to the dual use of IP addresses as "locators" (routing
labels) and "identifiers" (endpoint, or host, identifiers).  In HIP, public cryptographic keys, of a public/private key pair, are used as host identifiers, to which higher-layer protocols are bound instead of an IP address.  By using public keys (and their representations) as host identifiers, dynamic changes to IP address sets can be directly authenticated between hosts, and if desired, strong authentication between hosts at the TCP/IP stack level can be obtained.

rfc/text/advanced/ap_35_authentication_cryptosign.md
oberstet: I will add a background section, and the two options: A) having an single operator controlled trustroot (a root CA self-signed cert), and B) having a shared trustroot that is stored on-chain

oberstet: Done

oberstet:
rgd the "Remote Attestation" feature: besides storing keys in a hardware security device (hinted at in the diagram for certificate chains), this is the 2nd feature related to hardware security .. at least as it is usually implemented. When a client authenticates, the router wants to establish trust into the authenticating device. Checking for the correct signature of the signed challenge is one aspect, checking the integrity of the client itself is another. "Remote Attestation" is about the latter. Here is the new section:

grafik

rfc/text/advanced/ap_35_authentication_cryptosign.md
```diff
@@ -1,3 +1,1124 @@
...
+
+{align="left"}
+```javascript
+[1, "realm1", {
+    "roles": {/* see below */},
+    "authmethods": ["cryptosign"],
+    "authid": "client01@example.com",
+    "authextra": {
+        "pubkey": "545efb0a2192db8d43f118e9bf9aee081466e1ef36c708b96ee6f62dddad9122",
+        "channel_binding": null
```
KSDaemon, line 71: As `channel_binding` is optional and is not filled in this example, I suggest omitting it completely for the clearance.

rfc/text/advanced/ap_35_authentication_cryptosign.md
```diff
@@ -1,3 +1,1124 @@
...
+
+The `HELLO.Details.authextra|dict` contains the following members for WAMP-Cryptosign:
+
+| Field             | Type     | Required | Description                                                                          |
+|-------------------|----------|----------|--------------------------------------------------------------------------------------|
+| pubkey            | string   | yes      | The client public key (32 bytes) as a Hex encoded string, e.g. `545efb0a2192db8d43f118e9bf9aee081466e1ef36c708b96ee6f62dddad9122` |
+| channel_binding   | string   | no       | If TLS channel binding is in use, the TLS channel binding type, e.g. `"tls-unique"`. |
+| challenge         | string   | no       | A client chosen, random challenge (32 bytes) as a Hex encoded string, to be signed by the router. |
+| trustroot         | string   | no       | When the client includes a client certificate (see below), the Ethereum address of the trustroot of the certificate, e.g. `0x72b3486d38E9f49215b487CeAaDF27D6acf22115` |
```
KSDaemon, line 87: There can be different options for `trustroot` field, for example `Standalone Trustroot`. But that is described later in the specification. So new spec readers will first reach this section. And i think it's a good idea to provide or describe in short all possible options here.

rfc/text/advanced/ap_35_authentication_cryptosign.md
```diff
@@ -1,3 +1,1124 @@
...
+1. *Open*
+2. *Permitted*
+3. *Private*
+
+With an *Open On-chain Trustroot*, new certificates can be added to a certificate chain freely
+and only requires a signature by the respective intermediate CA issuer.
+
+##### Standalone Trustroots
+
+For a *Standalone Trustroot* the `trustroot` MUST be specified in `HELLO.Details.authextra.trustroot|string`
```
KSDaemon, line 595: I think, there is no description of what is the value for `trustroot` field?  Is it some hex encoded string of root certificate fingerprint or something else?
For on-chain trustroot — there is a clear description, that it is eth address on public blockchain where root certificate is stored.

KSDaemon (changes requested):
Great additions! 
I like new `Remote Attestation` section!

There are some minor points i've commented

oberstet:
thanks for the review! I've revised the text.

fwiw, I've also added a new section in the intro .. not sure, maybe that could also be at the very end, kinda reference info. anyways, we can move it if you like!

grafik

rfc/text/advanced/ap_35_authentication_cryptosign.md
```diff
@@ -23,6 +23,35 @@ The following sections describe each of above features of *WAMP-Cryptosign*.
 Examples of complete authentication message exchanges can be found at the end of this
 chapter in [Example Message Exchanges](#examplemessageexchanges).
 
+In WAMP, the following **cryptographic primitives** are used with *WAMP-Cryptosign* authentication:
+
+**Elliptic Curves:**
+
+|  SECG        |  Usage in WAMP                                                           |
+|--------------|--------------------------------------------------------------------------|
+|  secp256r1   |  Transport Encryption (WAMP transport rncryption)                        |
```
om26er, line 32: minor typo here `rncryption`

KSDaemon:
> fwiw, I've also added a new section in the intro .. not sure, maybe that could also be at the very end, kinda reference info. anyways, we can move it if you like!

No, i like it and think that it is the right place for it! 
Smart security guys can simply skip it :) But others like me — definitely need some general words and links to used specifications.

KSDaemon:
I think this should be merged into spec. Any additions can be done later.

KSDaemon (approved):

Issue #167, opened 2015-06-25
Progressive Calls


oberstet:
We have "progressive CALL _results_" (https://github.com/tavendo/WAMP/blob/master/spec/advanced.md#progressive-call-results) in the AP.

With this feature, a Callee can produce results in pieces.

It would be consequent and useful in certain situations to allow Callers to issue a call progressively.

That is, send multiple CALL messages with same request ID, with a `CALL.Options.progress|bool := true` option, finalized by a last CALL message with no progress option set.

jedesah:
We have a use case that would greatly benefit from such an addition to the WAMP protocol.

We are implementing a voice search service. Processing voice is fairly slow and in order to minimize the latency experienced by the user of a voice enabled device, it is preferable to start streaming the audio as soon as the user starts speaking as opposed to only once all the audio has been recorded.

An ideal scenario is a WAMP client capable of initiating a progressive call and receiving a progressive result. This way it would be possible to implement a low-latency voice search service that can even return results before the user has done speaking.

oberstet:
Ah, interesting! Yeah, that's probably a good use case for us to verify the feature design.

I guess you will want the progress chunking in Caller->Callee and Callee->Caller directions allowed to be independent? That is, not only have 1 progressive CALL_RESULT exactly per 1 progressive CALL.

oberstet:
Sidenote: we need to make sure this feature works with "shared registrations". E.g. when the same procedure has been registered by multiple Callees, the very first progressive CALL message will make the router select a concrete Callee to forward *this first, and all subsequent CALL messages* to that selected Callee.

jedesah:
@oberstet Yes, being able to independently stream both to and from the client would be ideal.

jedesah:
> to forward this first, and all subsequent CALL messages to that selected Callee.

Yes! This is the most complicated issue I am dealing with implementing my own ad-hoc solution.

oberstet:
Yeah. Actually, in Crossbar.io at least, that's not hard, since we create a proper, stateful call object in the router to correlate messages that come back from the callee to the original call anyway.

jedesah:
Let me know if there is anything I can do to help get this in as fast as possible?

konsultaner:
[...]

PR #414, opened 2022-07-25
RPC progressive calls

KSDaemon:
Here is a draft with description of progressive call.

### DIFF
--- a/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
+++ b/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
@@ -332,7 +332,7 @@ The message flow for progressive results cancellation involves:
 
 Note: Any `ERROR` returned by the *Callee*, in response to the `INTERRUPT`, is ignored (same as in call canceling when mode="killnowait"). So, it is not necessary for the *Callee* to send an `ERROR` message.
 
-**Ignoring Progressive Call Requests**
+**Ignoring Progressive Call Results Requests**
 
 A *Callee* that does not support progressive results SHOULD ignore any `INVOCATION.Details.receive_progress` flag.
 
--- a/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
+++ b/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
@@ -332,7 +332,7 @@ The message flow for progressive results cancellation involves:
 
 Note: Any `ERROR` returned by the *Callee*, in response to the `INTERRUPT`, is ignored (same as in call canceling when mode="killnowait"). So, it is not necessary for the *Callee* to send an `ERROR` message.
 
-**Ignoring Progressive Call Results Requests**
+**Ignoring Requests for Progressive Call Results**
 
 A *Callee* that does not support progressive results SHOULD ignore any `INVOCATION.Details.receive_progress` flag.
 
--- a/rfc/text/advanced/ap_10_rpc_progressive_calls.md
+++ b/rfc/text/advanced/ap_10_rpc_progressive_calls.md
@@ -173,3 +173,58 @@ If binding all ongoing `progressive` calls to the same *Callee* is not required,
 
 If `CALL.Options.sticky` is not specified it is treated like `TRUE`, so all `progressive`
 calls go to the same *Callee*.
+
+**Progressive Call Cancellation**
+
+If the original *Caller* is no longer available for some reason (has left the realm), then the *Dealer* will send 
+an `INTERRUPT` to the *Callee*. The `INTERRUPT` will have `Options.mode` set to `"killnowait"` to indicate to the 
+client that no response should be sent to the `INTERRUPT`.
+
+```
+[INTERRUPT, INVOCATION.Request|id, Options|dict]
+```
+Options:
+```
+INTERRUPT.Options.mode|string == "killnowait"
+```
+
+Progressive call cancellation, like progressive call result cancellation closes an important safety gap: 
+In cases where progressive calls are used to stream data from *Callers*, and network connectivity is unreliable, 
+*Callers* my often get disconnected in the middle of sending progressive data. This can lead to unneeded memory
+consumption on *Dealer* and *Callee* sides, because of storing runtime data about ongoing calls.
+
+The message flow for progressive results cancellation involves:
+
+{align="left"}
+,------.            ,------.                 ,------.
+|Caller|            |Dealer|                 |Callee|
+`--+---'            `--+---'                 `--+---'
+   |  CALL (progress)  |                        |
+   | ----------------->|                        |
+   |                   | INVOCATION (progress)  |
+   |                   | ---------------------->|
+   |  CALL (progress)  |                        |
+   | ----------------->|                        |
+   |                   | INVOCATION (progress)  |
+   |                   | ---------------------->|
+,--+---.               |                        |
+|Caller|               |                        |
+`------'               |                        |
+ (gone)                |                        |
+                       |       INTERRUPT        |
+                       | ---------------------->|
+                       |                        |
+                    ,--+---.                 ,--+---.
+                    |Dealer|                 |Callee|
+                    `------'                 `------'
+
+
+Note: Any `ERROR` returned by the *Callee*, in response to the `INTERRUPT`, is ignored (same as in regular call 
+canceling when mode="killnowait"). So, it is not necessary for the *Callee* to send an `ERROR` message.
+
+**Ignoring Progressive Call Requests**
+
+A *Callee* that does not support progressive call SHOULD ignore any `INVOCATION.Details.progress` flag.
+
+A *Callee* that supports progressive calls, but does not support call canceling is considered by the *Dealer* to 
+not support progressive calls.
--- a/rfc/text/advanced/ap_10_rpc_progressive_calls.md
+++ b/rfc/text/advanced/ap_10_rpc_progressive_calls.md
@@ -2,8 +2,10 @@
 
 A *Caller* may issue a progressive call. This can be useful in few cases:
 
-* Payload is too big to send it whole in one request.
+* Payload is too big to send it whole in one request. E.g. uploading a file.
 * Payload is long term-based, and it is required to start processing it earlier.
+* Partial results should be made available to the caller as quickly as possible. This can be accomplished
+  in conjunction with progressive call results feature.
 * RPC is called too often. So overall processing can be optimized: there is no need to generate new id for request,
   initiate structures for new call, etc.
 
@@ -15,14 +17,29 @@ This results in efficient two-way streams between *Caller* and *Callee*.
 
 **Feature Announcement**
 
-Support for this advanced feature MUST be announced by *Callers* (`role := "caller"`), *Callees* (`role := "callee"`) and *Dealers* (`role := "dealer"`) via
+Support for this advanced feature MUST be announced by *Callers* (`role := "caller"`), *Callees* (`role := "callee"`) 
+and *Dealers* (`role := "dealer"`) via
 
 {align="left"}
 HELLO.Details.roles.<role>.features.progressive_calls|bool := true
 
+Progressive Calls can work only if all three nodes support and announced this feature. 
+
+Cases when *Caller* sends `CALL` message with `progressive_calls := true` without announcing it during `HELLO` 
+handshake MUST be treated as *PROTOCOL ERRORS* and underlying WAMP connections must be aborted with 
+`wamp.error.protocol_violation` error reason.
+
+Cases when *Caller* sends `CALL` message with `progressive_calls := true` to *Dealer*, that did not announce 
+progressive calls support during `WELCOME` handshake MUST be treated as *PROTOCOL ERRORS* and underlying WAMP 
+connections must be aborted with `wamp.error.protocol_violation` error reason.
+
+Cases when *Caller* sends `CALL` message with `progressive_calls := true` to the *Dealer* that supports this feature,
+which then must be routed to the *Callee* which doesn't support progressive calls MUST be treated as *APPLICATION ERRORS*
+and *Dealer* MUST respond to *Caller* with `wamp.error.feature_not_supported` error message.
+
 **Message Flow**
 
-The message flow for progressive call when *Callee* waits for all chunks before processing:
+The message flow for progressive call when *Callee* waits for all chunks before processing and sending single result:
 
 {align="left"}
 ,------.            ,------.                 ,------.
@@ -224,7 +241,9 @@ canceling when mode="killnowait"). So, it is not necessary for the *Callee* to s
 
 **Ignoring Progressive Call Requests**
 
-A *Callee* that does not support progressive call SHOULD ignore any `INVOCATION.Details.progress` flag.
+Unlike other advanced features, this one can not be omitted by *Callee*.
+So if *Callee* doesn't support this feature, *Dealer* MUST respond to *Caller* with 
+`wamp.error.feature_not_supported` error message. 
 
 A *Callee* that supports progressive calls, but does not support call canceling is considered by the *Dealer* to 
 not support progressive calls.
--- a/rfc/text/advanced/ap_10_rpc_progressive_calls.md
+++ b/rfc/text/advanced/ap_10_rpc_progressive_calls.md
@@ -5,7 +5,7 @@ A *Caller* may issue a progressive call. This can be useful in few cases:
 * Payload is too big to send it whole in one request. E.g. uploading a file.
 * Payload is long term-based, and it is required to start processing it earlier.
 * Partial results should be made available to the caller as quickly as possible. This can be accomplished
-  in conjunction with progressive call results feature.
+  in conjunction with Progressive Call Results feature.
 * RPC is called too often. So overall processing can be optimized: there is no need to generate new id for request,
   initiate structures for new call, etc.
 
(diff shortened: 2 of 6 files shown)


rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,230 @@
...
+   |                   |                        |
+   |  RESULT or ERROR  |                        |
+   | <-----------------|                        |
+,--+---.            ,--+---.                 ,--+---.
+|Caller|            |Dealer|                 |Callee|
+`------'            `------'                 `------'
+
+As `progressive` call is still the same call *Callee* must send the same `Request|id` with every `CALL` and *Dealer*
+must also send the same `Request|id` with every `INVOCATION` to the *Callee*.
+
```
oberstet, line 58: Above diagram illustrates _one_ case for progressive calls. There are multiple more. There are these top level cases:

i) non-progressive call + non-progressive call results => a simple WAMP call
ii) non-progressive call + progressive call results
iii) progressive call + non-progressive call results => above message diagram
iv) progressive call + progressive call results => a bidirectionally streaming call

For iv), the crucial question we need to answer first is: what is the relation between individual CALLs (with progress) to individual INVOCATIONs, YIELDs and RESULTs. If any. The weakest restriction would be:

* the last CALL (final) must be forwarded in (a last) INVOCATION (final), and must be answered by RESULT (final), after which no more messages are valid for the (now completed) call

Sidenote: we need the progress flag on YIELD as well, but NOT on ERROR (there can be at most 1 ERROR for the whole call ... and ERROR ends the call)

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,230 @@
...
+    ]
+```
+
+A call invocation MUST *always* end in a *normal* `CALL` without `"progress": true` option, or explicitly set `"progress": false` which is default.
+
+**Progressive calls and shared registration**
+
+RPCs can have a multiple registrations (see `Shared Registration` feature) with different `<invocation_policies>`.
+In this case `progressive` `CALLs` can be routed to different *Callees*, what can lead to unexpected results.
+To bind `INVOCATIONs` to the same *Callee* *Caller* can specify `sticky` option during `CALL`
```
oberstet, line 163: This is a very powerful feature. Practically relevant, and very hard to do manually. Very good!

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,230 @@
...
+                    |Dealer|                 |Callee|
+                    `------'                 `------'
+
+
+Note: Any `ERROR` returned by the *Callee*, in response to the `INTERRUPT`, is ignored (same as in regular call 
+canceling when mode="killnowait"). So, it is not necessary for the *Callee* to send an `ERROR` message.
+
+**Ignoring Progressive Call Requests**
+
+A *Callee* that does not support progressive call SHOULD ignore any `INVOCATION.Details.progress` flag.
```
oberstet, line 227: IMO: No, it should kill the connection to the non-compliant router. If either the CALLEE or the DEALER did not announce the respective capability, and the other peer does use the feature anyways, this is violating the agreed protocol behavior, and thus the peer is bogus. Any subsequent talking to a bogus implementation is a security risk, and its semantics are undefined.

oberstet: in general, I am thinking like this: a single call involves 2 sessions (for the 2 connection legs CALLER-DEALER and CALLEE-DEALER), and there are 2 categories of problems:

a) a problem on a single leg, like using progress when it wasn't announced by both peers
b) a problem because one leg uses progress while the other doesn't support it

a) should are protocol errors, but b) is more an application error.

oberstet (changes requested):

KSDaemon:
@oberstet thnx for great points! I'll make an update!

rfc/text/advanced/ap_10_rpc_progressive_calls.md
KSDaemon: I have 3 and 4 cases schemes here. 1st is described in basic profile call and 2nd in progressive call results. So i think there is no need to copy paste them here.
And yep, we already have a `progress` flag in YIELD message, it's described also in progressive call results :)

oberstet (commented):
For me, this looks very good! Apart from 2 points:

1) the "Ignoring Progressive Call Results Requests" section: the callee must fail the connection since the router misbehaved (it did not honor the announcement of the callee that it doesn't support progressive calls / results)

2) IMO it would make things more explicit to use `YIELD (progress)` _and_ `YIELD (final)` to mirror how messages are annotated for the other message types

rfc/text/advanced/ap_09_rpc_progressive_call_results.md
```diff
@@ -332,7 +332,7 @@ The message flow for progressive results cancellation involves:
 
 Note: Any `ERROR` returned by the *Callee*, in response to the `INTERRUPT`, is ignored (same as in call canceling when mode="killnowait"). So, it is not necessary for the *Callee* to send an `ERROR` message.
 
-**Ignoring Progressive Call Requests**
+**Ignoring Progressive Call Results Requests**
```
ecorm, line 335: Double plural "Results Requests" doesn't roll off the tongue. May I suggest "Ignoring Requests for Progressive Call Results"?

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
 ## Progressive Calls {#rpc-progressive-calls}
 
-A procedure implemented by a *Callee* and registered at a *Dealer* may receive a progressive call. Progressive results can e.g. be used to start processing initial data where a larger data set may not yet have been generated or received by the *Caller*.
+A *Caller* may issue a progressive call. This can be useful in few cases:
 
-See this GitHub issue for more discussion: https://github.com/wamp-proto/wamp-proto/issues/167
+* Payload is too big to send it whole in one request. E.g. uploading a file.
+* Payload is long term-based, and it is required to start processing it earlier.
+* Partial results should be made available to the caller as quickly as possible. This can be accomplished
+  in conjunction with progressive call results feature.
```
ecorm, line 8: with the Progressive Call Results feature.

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+
+Support for this advanced feature MUST be announced by *Callers* (`role := "caller"`), *Callees* (`role := "callee"`) 
+and *Dealers* (`role := "dealer"`) via
+
+{align="left"}
+HELLO.Details.roles.<role>.features.progressive_calls|bool := true
+
+Progressive Calls can work only if all three nodes support and announced this feature. 
+
+Cases when *Caller* sends `CALL` message with `progressive_calls := true` without announcing it during `HELLO`
```
ecorm, line 28: Cases **where**

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+progressive calls support during `WELCOME` handshake MUST be treated as *PROTOCOL ERRORS* and underlying WAMP 
+connections must be aborted with `wamp.error.protocol_violation` error reason.
+
+Cases when *Caller* sends `CALL` message with `progressive_calls := true` to the *Dealer* that supports this feature,
+which then must be routed to the *Callee* which doesn't support progressive calls MUST be treated as *APPLICATION ERRORS*
+and *Dealer* MUST respond to *Caller* with `wamp.error.feature_not_supported` error message.
+
+**Message Flow**
+
+The message flow for progressive call when *Callee* waits for all chunks before processing and sending single result:
```
ecorm, line 42: for **a** progressive call

ecorm: sending **a** single result:

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+   |                   |  YIELD or ERROR        |
+   |                   | <----------------------|
+   |                   |                        |
+   |  RESULT or ERROR  |                        |
+   | <-----------------|                        |
+,--+---.            ,--+---.                 ,--+---.
+|Caller|            |Dealer|                 |Callee|
+`------'            `------'                 `------'
+
+As `progressive` call is still the same call *Callee* must send the same `Request|id` with every `CALL` and *Dealer*
```
ecorm, line 73: May I suggest: "All `CALL`, `INVOCATION`, `YIELD`, `RESULT` and `ERROR` messages belonging to the same overall progressive call MUST use the same `Request|Id`".

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+must also send the same `Request|id` with every `INVOCATION` to the *Callee*.
+
+
+The message flow for progressive call when *Callee* starts sending progressive call results immediately. Please note,
```
ecorm, line 77: The following message flow illustrates a progressive call when...

ecorm: Please note that (no comma)

ecorm: `YIELD` messages **don't**

ecorm: don't need to be **matched**

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+        77245,
+        {
+            "progress": true
+        },
+        "com.myapp.get_country_by_coords",
+        [50.450001, 30.523333]
+    ]
+```
+
+If the *Callee* supports progressive calls, the *Dealer* will forward the *Caller's* willingness to send progressive calls by setting
```
ecorm, line 153: will -> shall

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+        35224,
+        379,
+        {
+            "progress": true
+        },
+        [50.450001, 30.523333]
+    ]
+```
+
+A call invocation MUST *always* end in a *normal* `CALL` without `"progress": true` option, or explicitly set `"progress": false` which is default.
```
ecorm, line 174: is **the** default

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+
+**Progressive calls and shared registration**
+
+RPCs can have a multiple registrations (see `Shared Registration` feature) with different `<invocation_policies>`.
+In this case `progressive` `CALLs` can be routed to different *Callees*, what can lead to unexpected results.
```
ecorm, line 179: **which** can lead to

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+To bind `INVOCATIONs` to the same *Callee* *Caller* can specify `sticky` option during `CALL`
```
ecorm, line 180: *Callee*, **the** *Caller* can specify **the** `sticky`

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+
+{align="left"}
+CALL.Options.sticky|bool := true
+
+
+In this case *Dealer* must make a first `INVOCATION` based on `<invocation_policy>` and then route all next `progressive`
```
ecorm, line 186: **the** Dealer

ecorm: all next -> all subsequent

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+To bind `INVOCATIONs` to the same *Callee* *Caller* can specify `sticky` option during `CALL`
+
+{align="left"}
+CALL.Options.sticky|bool := true
+
+
+In this case *Dealer* must make a first `INVOCATION` based on `<invocation_policy>` and then route all next `progressive` 
+calls to the same *Callee*.
+
+If binding all ongoing `progressive` calls to the same *Callee* is not required, *Caller* can set `sticky` option to `FALSE`.
```
ecorm, line 189: **the** sticky option

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+{align="left"}
+CALL.Options.sticky|bool := true
+
+
+In this case *Dealer* must make a first `INVOCATION` based on `<invocation_policy>` and then route all next `progressive` 
+calls to the same *Callee*.
+
+If binding all ongoing `progressive` calls to the same *Callee* is not required, *Caller* can set `sticky` option to `FALSE`. 
+
+If `CALL.Options.sticky` is not specified it is treated like `TRUE`, so all `progressive`
```
ecorm, line 191: Missing comma-> specified, it is

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+
+```
+[INTERRUPT, INVOCATION.Request|id, Options|dict]
+```
+Options:
+```
+INTERRUPT.Options.mode|string == "killnowait"
+```
+
+Progressive call cancellation, like progressive call result cancellation closes an important safety gap:
```
ecorm, line 208: progressive call result cancellation -> progressive call results cancellation

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+```
+[INTERRUPT, INVOCATION.Request|id, Options|dict]
+```
+Options:
+```
+INTERRUPT.Options.mode|string == "killnowait"
+```
+
+Progressive call cancellation, like progressive call result cancellation closes an important safety gap: 
+In cases where progressive calls are used to stream data from *Callers*, and network connectivity is unreliable,
```
ecorm, line 209: from **a Caller** (Plural Callers is confusing as it implies multiple callers involved in the same RPC)

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+[INTERRUPT, INVOCATION.Request|id, Options|dict]
+```
+Options:
+```
+INTERRUPT.Options.mode|string == "killnowait"
+```
+
+Progressive call cancellation, like progressive call result cancellation closes an important safety gap: 
+In cases where progressive calls are used to stream data from *Callers*, and network connectivity is unreliable, 
+*Callers* my often get disconnected in the middle of sending progressive data. This can lead to unneeded memory
```
ecorm, line 210: a Caller may

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+consumption on *Dealer* and *Callee* sides, because of storing runtime data about ongoing calls.
```
ecorm, line 211: for the Dealer and Callee, due to the need to store temporary metadata about ongoing calls.

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+
+The message flow for progressive results cancellation involves:
```
ecorm, line 213: The message flow for cancelling progressive calls involves:

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+ (gone)                |                        |
+                       |       INTERRUPT        |
+                       | ---------------------->|
+                       |                        |
+                    ,--+---.                 ,--+---.
+                    |Dealer|                 |Callee|
+                    `------'                 `------'
+
+
+Note: Any `ERROR` returned by the *Callee*, in response to the `INTERRUPT`, is ignored (same as in regular call
```
ecorm, line 239: in response to **an** INTERRUPT

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+                    |Dealer|                 |Callee|
+                    `------'                 `------'
+
+
+Note: Any `ERROR` returned by the *Callee*, in response to the `INTERRUPT`, is ignored (same as in regular call 
+canceling when mode="killnowait"). So, it is not necessary for the *Callee* to send an `ERROR` message.
+
+**Ignoring Progressive Call Requests**
+
+Unlike other advanced features, this one can not be omitted by *Callee*.
```
ecorm, line 244: Unlike other advanced features, the Callee* cannot be unaware of progressive call requests.

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+So if *Callee* doesn't support this feature, *Dealer* MUST respond to *Caller* with
```
ecorm, line 245: Therefore, if a Callee

ecorm: **a** Dealer MUST respond to *Caller* with **an** `wamp.error.feature_not_supported` error message.

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,249 @@
...
+Note: Any `ERROR` returned by the *Callee*, in response to the `INTERRUPT`, is ignored (same as in regular call 
+canceling when mode="killnowait"). So, it is not necessary for the *Callee* to send an `ERROR` message.
+
+**Ignoring Progressive Call Requests**
+
+Unlike other advanced features, this one can not be omitted by *Callee*.
+So if *Callee* doesn't support this feature, *Dealer* MUST respond to *Caller* with 
+`wamp.error.feature_not_supported` error message. 
+
+A *Callee* that supports progressive calls, but does not support call canceling is considered by the *Dealer* to
```
ecorm, line 248: but does not support call cancelling, shall be considered by the *Dealer* as not supporting progressive calls

rfc/text/advanced/ap_48_uris.md
```diff
@@ -31,3 +31,9 @@ A *Dealer* could not perform a call, since a procedure with the given URI is reg
 
 {align="left"}
         wamp.error.no_available_callee
+
+A *Dealer* received a `CALL` message with advanced features that can not be processed by *Callee*
```
ecorm, line 35: can not -> cannot

ecorm: by **the** Callee

ecorm (changes requested):
Grammar fixes and suggestions for better readabilty

rfc/text/advanced/ap_09_rpc_progressive_call_results.md
KSDaemon: +1

rfc/text/advanced/ap_10_rpc_progressive_calls.md
KSDaemon: thnx!

rfc/text/advanced/ap_10_rpc_progressive_calls.md
KSDaemon: thnx!

rfc/text/advanced/ap_10_rpc_progressive_calls.md
KSDaemon: thnx! 
I always miss articles :((((

KSDaemon:
> 1. the "Ignoring Progressive Call Results Requests" section: the callee must fail the connection since the router misbehaved (it did not honor the announcement of the callee that it doesn't support progressive calls / results)

Well... Progressive call can not be figured out until it is issued. So when callee registers RPC — everything was ok. Even after that if RPC is called normally (without progress = true) — it is still ok. Caller doesn't know anything about capabilities of caller. The only time when dealer faces this problem — is when it receives CALL message with `progress=true` but it knows that caller doesn't support this feature. 

I think it is an application level error rather than protocol.

rfc/text/advanced/ap_10_rpc_progressive_calls.md
KSDaemon: indeed. Thnx!

oberstet:
one more formatting problem with diagrams: missing whitespace prefix:
Bildschirmfoto von 2022-07-26 01-37-46
Bildschirmfoto von 2022-07-26 01-37-10

rfc/text/advanced/ap_10_rpc_progressive_calls.md
KSDaemon: Hm... Is it not confusing? I mean `Request|Id` is different between CALL — INVOCATION — YIELD — RESULT. So it should be the same but within a related message type.

KSDaemon:
> one more formatting problem with diagrams: missing whitespace prefix: Bildschirmfoto von 2022-07-26 01-37-46 Bildschirmfoto von 2022-07-26 01-37-10

thnx! Fixed that also!

KSDaemon:
@oberstet @ecorm Thank you for ur review comments! I fixed most of them. Please have a [i hope final] look :)

KSDaemon:
Снимок экрана 2022-07-26 в 13 45 41

oberstet:
> Well... Progressive call can not be figured out until it is issued. ... I think it is an application level error rather than protocol.

Right, agreed! I was confused;) It needs to be an application error issued by the dealer.

oberstet (approved):

KSDaemon:
This will close #167

KSDaemon:
OOooookay... If no one else minds — I'll finally merge it! There is still one point by @ecorm. But we can fix it quickly later. 
Thnx everyone who was participating in the review! Together we made a great job!

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -1,5 +1,250 @@
...
+     ,------.                  ,------.                 ,------.
+     |Caller|                  |Dealer|                 |Callee|
+     `--+---'                  `--+---'                 `--+---'
+        |     CALL (progress)     |                        |
+        |------------------------>|                        |
+        |                         | INVOCATION (progress)  |
+        |                         |----------------------->|
+        |                         |                        |
+        |                         |   YIELD (progress)     |
+        |                         |<-----------------------|
```
bboozzoo, line 91: Is it possible to return ERROR early if the caller knows that the request cannot be processed?

KSDaemon: Yes, of course! If for some reason `Callee` understand that it can not process incoming `INVOCATION` — it can immediately send `ERROR` message. Thats ok.

oberstet: Yep, and should `Callee` decide to answer with `ERROR`, that concludes that invocation (no more messages from the `Callee`).

Btw: did we discuss combination of progressive calls & shared registrations for such a case already? What I mean is: assume there are 10 Callees with a shared registration of procedure P, and one of the Callees returns an ERROR while the progressive call is still ongoing, the *Router* MUST interrupt all other 9 Callees (by sending them an `INTERRUPT`), _and_ immediately send back `ERROR` to the original Caller

KSDaemon: I think we should clarify this in spec... So i've created #418

PR #417, opened 2022-07-28
Payload Passthru Mode [WIP]

KSDaemon:
Describing and discussing payload transparency feature

### DIFF
--- a/rfc/text/advanced/ap_50_payload_passthru_mode.md
+++ b/rfc/text/advanced/ap_50_payload_passthru_mode.md
@@ -3,18 +3,18 @@
 In some situations, you may want to reduce the access the router has to the information users transmit, or
 payload data is presented in some specific format that can not be simply recognized by WAMP router serializer.
 
-These are example use cases:
+Here are some use cases:
 
-* Using WAMP alongside with gateways to other technologies like MQTT Brokers or AMQP Queues. So actual payload 
-  is for example mqtt message that should be simply delivered to wamp topic as is. 
-* Sensitive user data that should be delivered to target *Callee* without any possibility to unveil it during delivery.
+* Using WAMP via gateways to other technologies like MQTT Brokers or AMQP Queues. So the actual payload 
+  is, for example, MQTT message that should be delivered to a WAMP topic as is. 
+* Sensitive user data that should be delivered to a target *Callee* without any possibility of unveiling it in transit.
 
-This needs can be covered with `Payload Passthru Mode` feature. This feature allows:
+The above use cases can be fulfilled with the `Payload Passthru Mode` feature. This feature allows:
 
-* Specifying additional attributes during `CALL`, `PUBLISH`, `EVENT`, `YIELD`, `RESULT` messages 
-  to signal *Router* to skip payload inspection.
-* Additional attributes are passed then alongside with `INVOCATION` and `ERROR` messages
-* Encrypting and decrypting payload with crypto algorithms.
+* Specifying additional attributes within `CALL`, `PUBLISH`, `EVENT`, `YIELD`, `RESULT` messages 
+  to signal the *Router* to skip payload inspection/conversion.
+* The forwarding of these additional attributes via `INVOCATION` and `ERROR` messages
+* Encrypting and decrypting payload using cryptographic algorithms.
 * Providing additional information about payload format and type.
 
 **Feature Announcement**
@@ -29,47 +29,49 @@ HELLO.Details.roles.<role>.features.payload_passthru_mode|bool := true
 Payload Passthru Mode can work only if all three nodes (*Caller*, *Dealer*, *Callee* or
 *Publisher*, *Broker*, *Subscriber*) support and announced this feature.
 
-Cases where *Caller* sends `CALL` message with `payload passthru` without announcing it during `HELLO`
+Cases where a *Caller* sends a `CALL` message with `payload passthru` without announcing it during the `HELLO`
 handshake MUST be treated as *PROTOCOL ERRORS* and underlying WAMP connections must be aborted with
-`wamp.error.protocol_violation` error reason.
+the `wamp.error.protocol_violation` error reason.
 
-Cases where *Caller* sends `CALL` message with `payload passthru` to *Dealer*, that did not announce
-`payload passthru` support during `WELCOME` handshake MUST be treated as *PROTOCOL ERRORS* and underlying WAMP
-connections must be aborted with `wamp.error.protocol_violation` error reason.
+Cases where a *Caller* sends a `CALL` message with `payload passthru` to a *Dealer*, the latter not announcing
+`payload passthru` support during `WELCOME` handshake MUST be treated as *PROTOCOL ERRORS* and the underlying WAMP
+connections must be aborted with the `wamp.error.protocol_violation` error reason.
 
-Cases where *Caller* sends `CALL` message with `payload passthru` to the *Dealer* that supports this feature,
-which then must be routed to the *Callee* which doesn't support `payload passthru` MUST be treated as 
-*APPLICATION ERRORS* and *Dealer* MUST respond to *Caller* with `wamp.error.feature_not_supported` error message.
+Cases where a *Caller* sends a `CALL` message with `payload passthru` to a *Dealer* that supports this feature,
+which then must be routed to a *Callee* which doesn't support `payload passthru`, MUST be treated as 
+*APPLICATION ERRORS* and the *Dealer* MUST respond to the *Caller* with a `wamp.error.feature_not_supported` 
+error message.
 
-Cases where *Publisher* sends `PUBLISH` message with `payload passthru` without announcing it during `HELLO`
-handshake MUST be treated as *PROTOCOL ERRORS* and underlying WAMP connections must be aborted with
-`wamp.error.protocol_violation` error reason.
+Cases where a *Publisher* sends a `PUBLISH` message with `payload passthru`, without announcing it during `HELLO`
+handshake, MUST be treated as *PROTOCOL ERRORS* and the underlying WAMP connections must be aborted with
+the `wamp.error.protocol_violation` error reason.
 
-Cases where *Publisher* sends `PUBLISH` message with `payload passthru` to *Broker*, that did not announce
-`payload passthru` support during `WELCOME` handshake MUST be treated as *PROTOCOL ERRORS* and underlying WAMP
-connections must be aborted with `wamp.error.protocol_violation` error reason.
+Cases where a *Publisher* sends a `PUBLISH` message with `payload passthru` to a *Broker*, with the latter not 
+announcing `payload passthru` support during the `WELCOME` handshake, MUST be treated as *PROTOCOL ERRORS* and 
+the underlying WAMP connections must be aborted with the `wamp.error.protocol_violation` error reason.
 
-Cases where *Publisher* sends `PUBLISH` message with `payload passthru` to the *Broker* that supports this feature,
-which then must be routed to the *Subscriber* which doesn't support `payload passthru` cannot be recognized at the
-protocol level due to asynchronous message processing and must be covered at *Subscriber* side.
+Cases where a *Publisher* sends a `PUBLISH` message with `payload passthru` to a *Broker* that supports this feature,
+which then must be routed to a *Subscriber* which doesn't support `payload passthru`, cannot be recognized at the
+protocol level due to asynchronous message processing and must be covered at the *Subscriber* side.
 
-Cases where *Callee* sends `YIELD` message with `payload passthru` without announcing it during `HELLO`
-handshake MUST be treated as *PROTOCOL ERRORS* and underlying WAMP connections must be aborted with
-`wamp.error.protocol_violation` error reason.
+Cases where a *Callee* sends a `YIELD` message with `payload passthru` without announcing it during the `HELLO`
+handshake MUST be treated as *PROTOCOL ERRORS* and the underlying WAMP connections must be aborted with
+the `wamp.error.protocol_violation` error reason.
 
-Cases where *Callee* sends `YIELD` message with `payload passthru` to *Dealer*, that did not announce
-`payload passthru` support during `WELCOME` handshake MUST be treated as *PROTOCOL ERRORS* and underlying WAMP
-connections must be aborted with `wamp.error.protocol_violation` error reason.
+Cases where a *Callee* sends a `YIELD` message with `payload passthru` to a *Dealer*, with the latter not announcing
+`payload passthru` support during the `WELCOME` handshake, MUST be treated as *PROTOCOL ERRORS* and the 
+underlying WAMP connections must be aborted with the `wamp.error.protocol_violation` error reason.
 
-Cases where *Callee* sends `YIELD` message with `payload passthru` to the *Dealer* that supports this feature,
-which then must be routed to the *Caller* which doesn't support `payload passthru` MUST be treated as
-*APPLICATION ERRORS* and *Dealer* MUST respond to *Callee* with `wamp.error.feature_not_supported` error message.
+Cases where a *Callee* sends a `YIELD` message with `payload passthru` to a *Dealer* that supports this feature,
+which then must be routed to the *Caller* which doesn't support `payload passthru`, MUST be treated as
+*APPLICATION ERRORS* and the *Dealer* MUST respond to the *Callee* with a `wamp.error.feature_not_supported` 
+error message.
 
-**Message attributes**
+**Message Attributes**
 
-For this feature to be in use, `CALL`, `PUBLISH` and `YIELD` messages options are extended with additional attributes. 
-This attributes then are passed as is within `INVOCATION`, `EVENT` and `RESULT` messages respectively and to `ERROR`
-message in case of failures.
+To use payload passthru mode, the options for `CALL`, `PUBLISH` and `YIELD` messages MUST be extended with additional 
+attributes. These additional attributes must be forwarded via `INVOCATION`, `EVENT` and `RESULT` messages, respectively, 
+as well as `ERROR` messages in the case of failures.
 
 {align="left"}
         CALL.Options.ppt_scheme|string
@@ -114,47 +116,49 @@ message in case of failures.
         ERROR.Options.ppt_keyid|string
 
 
-**ppt_scheme attribute**
+**ppt_scheme Attribute**
 
-`ppt_scheme` stands for Key Management Schema. It is required string attribute. This attribute can contain name or
-identifier of key management provider which is known to target peer, so it can be used to obtain information
-about encryption keys. *Router* understands that `Payload Passthru Mode` is in use
-by checking the existence and non-empty value of this attribute in `CALL`, `PUBLISH` and `YIELD` messages options.
+The `ppt_scheme` identifies the Key Management Schema. It is a required string attribute. This attribute can contain 
+the name or identifier of a key management provider that is known to the target peer, so it can be used to obtain 
+information about encryption keys. A *Router* can recognize that `Payload Passthru Mode` is in use
+by checking the existence and non-empty value of this attribute within the options of `CALL`, `PUBLISH` and `YIELD` 
+messages.
 
-**ppt_serializer attribute**
+**ppt_serializer Attribute**
 
-`ppt_serializer` attribute is optional. It specifies what serializer was used to build up payload object.
-It can be `mqtt`, `amqp`, `stomp` value just to inform that coming data is from related technologies or ordinary 
-`json`, `msgpack`, `cbor`, `flatbuffers` data serializers.
+The `ppt_serializer` attribute is optional. It specifies what serializer was used to encode the payload.
+It can be a value a such as `mqtt`, `amqp`, `stomp` to indicate that the incoming data is tunneling through such 
+technologies, or it can be ordinary `json`, `msgpack`, `cbor`, `flatbuffers` data serializers.
 
-**ppt_cipher attribute**
+**ppt_cipher Attribute**
 
-`ppt_cipher` attribute is optional. It is required if payload is encrypted. This attribute specifies cryptographic
-algorithm that was used to encrypt payload. It can be `xsalsa20poly1305`, `aes256gcm` for now.
+The `ppt_cipher` attribute is optional. It is required if the payload is encrypted. This attribute specifies the 
+cryptographic algorithm that was used to encrypt the payload. It can be `xsalsa20poly1305`, `aes256gcm` for now.
 
-**ppt_keyid attribute**
+**ppt_keyid Attribute**
 
-`ppt_keyid` attribute is optional. This attribute can contain encryption key id that was used to encrypt payload.
-`ppt_keyid` attribute is always a type of string. The value can be a hex-encoded string, uri, DNS name, 
-Ethereum address, UUID identifier - any meaningful value by which target peer can choose private key 
-without guessing. Format of the value may depend on `ppt_scheme` attribute.
+The `ppt_keyid` attribute is optional. This attribute can contain the encryption key id that was used to encrypt 
+the payload. The `ppt_keyid` attribute is a string type. The value can be a hex-encoded string, URI, DNS name, 
+Ethereum address, UUID identifier - any meaningful value which allows the target peer to choose a private key 
+without guessing. The format of the value may depend on the `ppt_scheme` attribute.
 
-**ppt_ predefined schemes**
+**ppt_ Predefined Schemes**
 
-| Attribute      | Required? | Predefined Scheme (mqtt)  | Predefined Scheme (End-to-End Encryption) | Custom Scheme Example |
+| Attribute      | Required? | Predefined Scheme (MQTT)  | Predefined Scheme (End-to-End Encryption) | Custom Scheme Example |
 |----------------|-----------|---------------------------|-------------------------------------------|-----------------------|
 | ppt_scheme     | Y         | mqtt                      | wamp.eth                                  | x_my_ppt              |
 | ppt_serializer | N         | json, msgpack, cbor, mqtt | cbor, flatbuffers                         | custom                |
 | ppt_cipher     | N         | -                         | xsalsa20poly1305, aes256gcm               | custom                |
 | ppt_keyid      | N         | -                         | *                                         | custom                |
 
-*: Least significant 20 bytes (160 bits) of SHA256 of public key (32 bytes) of data encryption key as HEX encoded 
-string with prefix `0x`, and upper/lower case letters encoding a checksum according to EIP55.
+*: The least significant 20 bytes (160 bits) of the SHA256 of the public key (32 bytes) of the data encryption key, 
+as a hex-encoded string with prefix `0x` and either uppercase/lowercase alphabetic characters, encoding a 
+checksum according to EIP55.
 
-**Message structure**
+**Message Structure**
 
-With `Payload Passthru Mode` in use message payload MUST BE sent as one binary item inside 
-`Arguments|list`, while `ArgumentsKw|dict` MUST BE missing or empty.
+When `Payload Passthru Mode` is in use, the message payload MUST be sent as one binary item within 
+`Arguments|list`, while `ArgumentsKw|dict` MUST be absent or empty.
 
 *Example.* Caller-to-Dealer `CALL` with encryption and key ID
 
@@ -176,7 +180,8 @@ With `Payload Passthru Mode` in use message payload MUST BE sent as one binary i
 
 *Example.* Caller-to-Dealer progressive `CALL` with encryption and key ID.
 
-Nothing prevents to use `Payload Passthru Mode` with other features, for example `Progressive Calls`
+Note that nothing prevents the use of `Payload Passthru Mode` with other features such as, 
+for example, `Progressive Calls`.
 
 {align="left"}
 ```json
@@ -195,8 +200,8 @@ Nothing prevents to use `Payload Passthru Mode` with other features, for example
     ]
 ```
 
-*Example.* Caller-to-Dealer `CALL` with mqtt payload. Specifying `"ppt_serializer": "mqtt"` means, that 
-original mqtt message payload is passed as WAMP payload message without any transcoding, as is.
+*Example.* Caller-to-Dealer `CALL` with MQTT payload. Specifying `"ppt_serializer": "mqtt"` means that 
+the original MQTT message payload is passed as WAMP payload message as is, without any transcoding, as is.
 
 {align="left"}
 ```json
@@ -212,9 +217,9 @@ original mqtt message payload is passed as WAMP payload message without any tran
     ]
 ```
 
-*Example.* Caller-to-Dealer `CALL` with mqtt payload. Specifying `"ppt_scheme": "mqtt"` simply indicates, that
-original source of payload data is received from related system. Specifying `"ppt_serializer": "json"` means, that
-original mqtt message payload was parsed and encoded with `json` serializer.
+*Example.* Caller-to-Dealer `CALL` with MQTT payload. Specifying `"ppt_scheme": "mqtt"` simply indicates that
+the original source of payload data is received from a related system. Specifying `"ppt_serializer": "json"` means that
+the original MQTT message payload was parsed and encoded with the `json` serializer.
 
 {align="left"}
 ```json
@@ -248,7 +253,7 @@ original mqtt message payload was parsed and encoded with `json` serializer.
     ]
 ```
 
-*Example.* Dealer-to-Callee `INVOCATION` with mqtt payload
+*Example.* Dealer-to-Callee `INVOCATION` with MQTT payload
 
 {align="left"}
 ```json
@@ -283,7 +288,7 @@ original mqtt message payload was parsed and encoded with `json` serializer.
 
 *Example.* Callee-to-Dealer progressive `YIELD` with encryption and key ID
 
-Nothing prevents to use `Payload Passthru Mode` with other features, for example `Progressive Call Results`
+Nothing prevents the use of `Payload Passthru Mode` with other features such as, for example, `Progressive Call Results`.
 
 {align="left"}
 ```json
@@ -320,7 +325,7 @@ Nothing prevents to use `Payload Passthru Mode` with other features, for example
 
 *Example.* Dealer-to-Caller progressive `RESULT` with encryption and key ID
 
-Nothing prevents to use `Payload Passthru Mode` with other features, for example `Progressive Call Results`
+Nothing prevents the use of `Payload Passthru Mode` with other features such as, for example, `Progressive Call Results`.
 
 {align="left"}
 ```json
@@ -393,9 +398,9 @@ Nothing prevents to use `Payload Passthru Mode` with other features, for example
     ]
 ```
 
-**About supported serializers and cryptographic ciphers**
+**About Supported Serializers and Cryptographic Ciphers**
 
-WAMP works as infrastructure for delivering messages between peers. Regardless of what encryption algorithm 
-and serializer were chosen for `Payload Passthru Mode`, *Router* will not inspect and analyze related encryption 
-message options and payload. It is up to an application side code responsibility to use serializers and 
-ciphers known to every peer involved into message processing.
+WAMP serves as infrastructure for delivering messages between peers. Regardless of what encryption algorithm 
+and serializer were chosen for `Payload Passthru Mode`, a *Router* shall not inspect and analyze the `ppt_` options 
+and payload of encrypted messages. The application is responsible for choosing serializers and 
+ciphers known to every peer involved in message processing.
--- a/rfc/text/advanced/ap_50_payload_passthru_mode.md
+++ b/rfc/text/advanced/ap_50_payload_passthru_mode.md
@@ -4,6 +4,7 @@ In some situations, you may want to reduce the access the router has to the info
 payload data is presented in some specific format that can not be simply recognized by WAMP router serializer.
 
 These are example use cases:
+
 * Using WAMP alongside with gateways to other technologies like MQTT Brokers or AMQP Queues. So actual payload 
   is for example mqtt message that should be simply delivered to wamp topic as is. 
 * Sensitive user data that should be delivered to target *Callee* without any possibility to unveil it during delivery.
@@ -12,6 +13,7 @@ This needs can be covered with `Payload Passthru Mode` feature. This feature all
 
 * Specifying additional attributes during `CALL`, `PUBLISH`, `EVENT`, `YIELD`, `RESULT` messages 
   to signal *Router* to skip payload inspection.
+* Additional attributes are passed then alongside with `INVOCATION` and `ERROR` messages
 * Encrypting and decrypting payload with crypto algorithms.
 * Providing additional information about payload format and type.
 
@@ -70,65 +72,65 @@ This attributes then are passed as is within `INVOCATION`, `EVENT` and `RESULT` 
 message in case of failures.
 
 {align="left"}
+        CALL.Options.ppt_scheme|string
         CALL.Options.ppt_serializer|string
         CALL.Options.ppt_cipher|string
-        CALL.Options.ppt_scheme|string
         CALL.Options.ppt_keyid|string
         ---
+        INVOCATION.Details.ppt_scheme|string
         INVOCATION.Details.ppt_serializer|string
         INVOCATION.Details.ppt_cipher|string
-        INVOCATION.Details.ppt_scheme|string
         INVOCATION.Details.ppt_keyid|string
         ---
+        YIELD.Options.ppt_scheme|string
         YIELD.Options.ppt_serializer|string
         YIELD.Options.ppt_cipher|string
-        YIELD.Options.ppt_scheme|string
         YIELD.Options.ppt_keyid|string
         ---
+        RESULT.Details.ppt_scheme|string
         RESULT.Details.ppt_serializer|string
         RESULT.Details.ppt_cipher|string
-        RESULT.Details.ppt_scheme|string
         RESULT.Details.ppt_keyid|string
         ---
+        ERROR.Details.ppt_scheme|string
         ERROR.Details.ppt_serializer|string
         ERROR.Details.ppt_cipher|string
-        ERROR.Details.ppt_scheme|string
         ERROR.Details.ppt_keyid|string
 
 {align="left"}
+        PUBLISH.Options.ppt_scheme|string
         PUBLISH.Options.ppt_serializer|string
         PUBLISH.Options.ppt_cipher|string
-        PUBLISH.Options.ppt_scheme|string
         PUBLISH.Options.ppt_keyid|string
         ---
+        EVENT.Details.ppt_scheme|string
         EVENT.Details.ppt_serializer|string
         EVENT.Details.ppt_cipher|string
-        EVENT.Details.ppt_scheme|string
         EVENT.Details.ppt_keyid|string
         ---
+        ERROR.Options.ppt_scheme|string
         ERROR.Options.ppt_serializer|string
         ERROR.Options.ppt_cipher|string
-        ERROR.Options.ppt_scheme|string
         ERROR.Options.ppt_keyid|string
 
 
+**ppt_scheme attribute**
+
+`ppt_scheme` stands for Key Management Schema. It is required string attribute. This attribute can contain name or
+identifier of key management provider which is known to target peer, so it can be used to obtain information
+about encryption keys. *Router* understands that `Payload Passthru Mode` is in use
+by checking the existence and non-empty value of this attribute in `CALL`, `PUBLISH` and `YIELD` messages options.
+
 **ppt_serializer attribute**
 
-`ppt_serializer` attribute is required. It specifies what serializer was used to build up payload object.
+`ppt_serializer` attribute is optional. It specifies what serializer was used to build up payload object.
 It can be `mqtt`, `amqp`, `stomp` value just to inform that coming data is from related technologies or ordinary 
-`json`, `msgpack`, `cbor`, `flatbuffers` data serializers. *Router* understands that `Payload Passthru Mode` is in use
-by checking the existence and non-empty value of this attribute in `CALL`, `PUBLISH` and `YIELD` messages options.
+`json`, `msgpack`, `cbor`, `flatbuffers` data serializers.
 
 **ppt_cipher attribute**
 
 `ppt_cipher` attribute is optional. It is required if payload is encrypted. This attribute specifies cryptographic
 algorithm that was used to encrypt payload. It can be `xsalsa20poly1305`, `aes256gcm` for now.
-
-**ppt_scheme attribute**
-
-`ppt_scheme` stands for Key Management Schema. It is optional string attribute. This attribute can contain name or
-identifier of key management provider which is known to target peer, so it can be used to obtain information
-about encryption keys.
 
 **ppt_keyid attribute**
 
@@ -136,6 +138,18 @@ about encryption keys.
 `ppt_keyid` attribute is always a type of string. The value can be a hex-encoded string, uri, DNS name, 
 Ethereum address, UUID identifier - any meaningful value by which target peer can choose private key 
 without guessing. Format of the value may depend on `ppt_scheme` attribute.
+
+**ppt_ predefined schemes**
+
+| Attribute      | Required? | Predefined Scheme (mqtt)  | Predefined Scheme (End-to-End Encryption) | Custom Scheme Example |
+|----------------|-----------|---------------------------|-------------------------------------------|-----------------------|
+| ppt_scheme     | Y         | mqtt                      | wamp.eth                                  | x_my_ppt              |
+| ppt_serializer | N         | json, msgpack, cbor, mqtt | cbor, flatbuffers                         | custom                |
+| ppt_cipher     | N         | -                         | xsalsa20poly1305, aes256gcm               | custom                |
+| ppt_keyid      | N         | -                         | *                                         | custom                |
+
+*: Least significant 20 bytes (160 bits) of SHA256 of public key (32 bytes) of data encryption key as HEX encoded 
+string with prefix “0x”, and upper/lower case letters encoding a checksum according to EIP55.
 
 **Message structure**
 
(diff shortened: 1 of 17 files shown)


rfc/text/advanced/ap_50_payload_transparency.md
```diff
@@ -0,0 +1,251 @@
+## Payload Transparency
```
ecorm, line 1: "Transparent" means that the contents of the payload is visible. The terminology you need to use is "opaque payload" or "payload opacity".

https://www.merriam-webster.com/dictionary/transparent
https://www.merriam-webster.com/dictionary/opaque
https://www.merriam-webster.com/dictionary/opacity
https://english.stackexchange.com/questions/87177/difference-between-opacity-and-opaqueness

oberstet:
@ecorm sorry for having this wrong term creep on! we need a better name

@KSDaemon apparently this term is wrong / suggests the wrong meaning, so we need a different name and feature ID - Emile mentioned this previously and raised my attention to this, but of course all the existing experimental code in AB and CB are still on the wrong term. sorry, you couldn't know.

"have the router forward app payload without touching or caring": what's a good name for that?

KSDaemon:
haha! Feeling so stupid))))) Really))) Transparency.... And we are discussing encryption%)))

oberstet:
rgd having a new field to distinguish between different types of `enc_keyid`:

* the value must a be string, always and for any of the subtypes
* the value could be: a Hex-encoded string, a UUID encoded in standard format, an Ethereum address, a forward (or reverse) DNS name, ... 
* which type a (base)-string-typed value does indeed have can be

1. auto-detected (eg https://github.com/crossbario/autobahn-python/blob/9f425ff5792348c2f6dc3e7c3297c82f7fb5e793/autobahn/wamp/message.py#L231)
2. determined by `enc_algo`
3. determined by a new attribute `enc_kms` ("key management schema")
4. determined by a prefix, eg `enc_keyid = "xbr://0xe59C7418403CF1D973485B36660728a5f4A8fF9c"`

oberstet:
rather than "payload transparency", how about one of these?

1. "raw payload mode"
2. "unprocessed payload mode"
3. "zero-touch payload mode"
4. "zero-parse payload"

KSDaemon:
What came to my mind:
* Payload shadowing or Shadow Payload
* Payload opacity by @ecorm 
* Payload hiding

oberstet:
@ecorm 

rgd "Payload opacity", doesn't that suggest there would be a level or grade of opacity? which isn't the case. as when the mode is on, the router doesn't care at all, and if it is off, the router fully parses according to the wamp session serializer

how about "Opaque Payload Mode"?

oberstet:
rgd adding a be attribute `enc_kms` or a prefix to identify the key ID / management scheme to be used (and in use for the ID string format):

if we do that, then we should redefine `enc_algo` as well

because currently, `enc_algo` determines _both_ the actual cryptographic encryption algorithm (eg XSalsa20Poly1305 for XBR) _and_ the naming schema in use (UUIDs mapped to ETH addresses) ...

IOW: what is the advantage of using 2 attributes to identify those? I mean, the client needs to know/determine both anyways

counter argument: right, but then why do we use a separate attribute `enc_serializer` and not only have exactly _one_ to determine all of:

* crypto algo used
* key management scheme used
* serializer used (before encryption)

mmmh;)

ecorm:
> "have the router forward app payload without touching or caring": what's a good name for that?

Perhaps "opaque" isn't the best term because the router can still see the payload (even if it can't understand it when encrypted).

How about:
- Intact Payloads
- Pristine Payloads
- Protected Payloads
- Payload Conservation
- Immaculate Payloads

KSDaemon:
I like `Protected Payloads`.

oberstet:
awesome!! *Protected Payloads* or *Protected Payload Mode* works perfectly for me as well (not a native speaker anyways, so I lack the real clue about connotations .. anyways;) +1

KSDaemon:
Cool! I'll update docs to → `Protected Payload Mode`

oberstet:
btw, the list of all WAMP message types that can have `enc_algo` and `args[0] == payload|binary`:

```
(base) oberstet@intel-nuci7:~$ grep "Payload|binary]" scm/crossbario/autobahn-python/autobahn/wamp/message.py 
    * ``[ERROR, REQUEST.Type|int, REQUEST.Request|id, Details|dict, Error|uri, Payload|binary]``
    * ``[PUBLISH, Request|id, Options|dict, Topic|uri, Payload|binary]``
    * ``[EVENT, SUBSCRIBED.Subscription|id, PUBLISHED.Publication|id, Details|dict, PUBLISH.Payload|binary]``
    * ``[CALL, Request|id, Options|dict, Procedure|uri, Payload|binary]``
    * ``[RESULT, CALL.Request|id, Details|dict, Payload|binary]``
    * ``[INVOCATION, Request|id, REGISTERED.Registration|id, Details|dict, Payload|binary]``
    * ``[YIELD, INVOCATION.Request|id, Options|dict, Payload|binary]``
```

ecorm:
While I was consulting the thesaurus, this scene from Lord of the Rings kept playing through my head: "Get back scum! The prisoners go back to Saruman alive and unspoiled."

oberstet:
Two more notes rgd feature announcement and consistent use thereafter:

What needs to be announced is probably not only *Protected Payload Mode*, but the specific variant, and in determine all of:
* crypto algo used, eg xsalsa20poly1305
* key management scheme used, eg eth
* serializer used (before encryption), eg cbor

And when a specific flavor is to be used, such as `eth.xsalsa20poly1305.cbor`, then **all** WAMP messages in that session must use the same?

If we want to relax that, then we need to state how this works .. eg at least, `RESULT.details.enc_algo == CALL.details.enc_algo` and similar must hold?

> fwiw, in AB/CB all demos/examples use one schema for everything ...

oberstet:
> Get back scum! The prisoners go back to Saruman alive and unspoiled.

=) sounds interesting! unfortunately, haven't seen lord of the rings, and internet search doesn't turn up the respective sequence. I will die ignorant;) or unspoiled? haha

oberstet:
to make the decision making / voting easier, let me try to summarize the two options we have rgd the attributes (as I understand what's been suggested that is):

A)

* `enc_serializer`, e.g. `"cbor"`: payload (before enc) serializer
* `enc_algo`, e.g. `"xsalsa20poly1305"`: payload cipher
* `enc_kms`, e.g `"eth"`: key management schema (which also determines the key ID string format)
* `enc_key`, e.g. `"0xe59C7418403CF1D973485B36660728a5f4A8fF9c"`: key ID according to format and semantics determined by `enc_kms`

OR

B)
* `enc_algo`: all 3 above combined, eg `eth.xsalsa20poly1305.cbor`
* `enc_key`: same as above

> fwiw, I'd be fine with both, and I would also be fine with renaming `enc_key` (as it is not the key itself, but an ID pointing to a key) ... it's just that it'll result in tedious work on AB/CB side;)

ecorm:
> =) sounds interesting! unfortunately, haven't seen lord of the rings, and internet search doesn't turn up the respective sequence. I will die ignorant;) or unspoiled? haha

https://www.youtube.com/watch?v=ufFOghMt1yI

oberstet:
> https://www.youtube.com/watch?v=ufFOghMt1yI

thanks!! so "unspoiled" is about the "meal" consisting of those poor guys. I see. interesting association while looking in the thesaurus rgd "opaque"=) fwiw, when I think of "opaque" and cinema, first thing that comes to my mind is andrei tarkowsky, eg stalker ... "the zone" is definitely opaque https://www.youtube.com/watch?v=YuOnfQd-aTw

om26er:
I guess we should update the title of this issue to be `WIP: Protected Payload Mode` ;-)

KSDaemon:
Ooookay! Next turn! 

* Renamed feature
* Return enc_key to original meaning
* Added enc_kms options
* Updated some descriptions
* Added info about supporting cyphers and serializers

2 all: Please have a look!

KSDaemon:
> to make the decision making / voting easier, let me try to summarize the two options we have rgd the attributes (as I understand what's been suggested that is):
> 
> A)
> B)
> C) as today in AB/CB

Well, I prefer A variant. It's more clear for end user to determine everything. In case of B) — we move some logic to users shoulders. I don't think that they want it :)

So I described A) in spec :)

Jopie64:
> haha! Feeling so stupid))))) Really))) Transparency.... And we are discussing encryption%)))

I thought it was called 'transparent' as in, the router passes the payload transparently. Like, it doesn't even see it, like a window...

oberstet:
> > haha! Feeling so stupid))))) Really))) Transparency.... And we are discussing encryption%)))
> 
> I thought it was called 'transparent' as in, the router passes the payload transparently. Like, it doesn't even see it, like a window...

In fact, both is true:

1. from a client point of view, the router treats the payload in a transparent manner .. as if it wasn't there at all (well, mostly, apart from size limits or such)
2. from a router point of view, the payload is treated opaquely .. it cannot look inside .. it (the payload from a router PoV) is opaque

So it is just a dual perspective.

Funny: I was thinking more like you .. from an app PoV ... but I'm fine with both. We just need to pick one.

KSDaemon:
> I thought it was called 'transparent' as in, the router passes the payload transparently. Like, it doesn't even see it, like a window...

Well, if in long it reads as `payload transparency to the router` — then why not!?)

But anyway, i think current chosen term is better!

rfc/text/advanced/ap_50_protected_payload_mode.md
```diff
@@ -0,0 +1,284 @@
...
+payload data is presented in some specific format that can not be simply recognized by WAMP router serializer.
+
+These are example use cases:
+* Using WAMP alongside with gateways to other technologies like MQTT Brokers or AMQP Queues. So actual payload 
+  is for example mqtt message that should be simply delivered to wamp topic as is. 
+* Sensitive user data that should be delivered to target *Callee* without any possibility to unveil it during delivery.
+
+This needs can be covered with `Protected Payload Mode` feature. This feature allows:
+
+* Specifying additional attributes during `CALL`, `PUBLISH`, `EVENT`, `YIELD`, `RESULT` messages
```
oberstet, line 13: we also need ERROR and INVOCATION

rfc/text/advanced/ap_50_protected_payload_mode.md
```diff
@@ -0,0 +1,284 @@
...
+* Providing additional information about payload format and type.
+
+**Feature Announcement**
+
+Support for this advanced feature MUST be announced by *Callers* (`role := "caller"`), *Callees* (`role := "callee"`),
+*Dealers* (`role := "dealer"`), *Publishers* (`role := "publisher"`), *Subscribers* (`role := "subscriber"`) 
+and *Brokers* (`role := "broker"`) via
+
+{align="left"}
+HELLO.Details.roles.<role>.features.protected_payload_mode|bool := true
```
oberstet, line 25: so this doesn't announce the specific way it will be used in the client, e.g. it doesn't say

```
protected_payload_mode = "eth.xsalsa20poly1305.cbor"
```

or

```
protected_payload_mode = "mqtt.native"
```
or

```
protected_payload_mode = ["eth.xsalsa20poly1305.cbor", "eth.xsalsa20poly1305.flatbuffers"]
```

oberstet (changes requested):

ecorm:
> I thought it was called 'transparent' as in, the router passes the payload transparently. Like, it doesn't even see it, like a window...

> In fact, both is true:
>
>    from a client point of view, the router treats the payload in a transparent manner .. as if it wasn't there at all (well, mostly, apart from size limits or such)
>    from a router point of view, the payload is treated opaquely .. it cannot look inside .. it (the payload from a router PoV) is opaque

"Transparent" is often misused to mean "invisible", as in a process that's hidden and done automatically. But "transparent" really means being able to see through it (like a jar made out of glass). The glass jar is not invisible, and you can see the details of what's inside it.

Consider "opaque types" in C/C++, in which the data structure details are completely hidden (you don't even know its size) and is passed around with a `void` pointer. Considering that "opaque" means the opposite of "transparent", then "transparent" doesn't make sense to describe a mechanism where details are ignored or hidden.

Check out this SO answer, with the highly upvoted comment that reflects my own opinion: https://stackoverflow.com/a/17385138/245265

ecorm:
While we're still bikeshedding the name of the feature, I just though of another alternative: "Payload Passthru" or "Payload Passthru Mode".

"Passthru" is used in digital audio to mean that the signal is passed through unchanged to the output.

meejah:
While scanning here (who doesn't love a nice "naming" bikeshed!) I was also going to suggest "payload pass-through" (maybe the "passthru" spelling is fine too) because:

- it doesn't _require_ encryption
- says precisely what it means (the server "passes through" the payload, without touching it)
- cases that _do_ use encryption can say "encrypted payload pass-through" for maximum clarity

oberstet:
"Payload Passthru": yeah, this is even better .. I also like the "Encrypted Payload Passthru" to specify the variants for that specific use case of it. @KSDaemon what's your take? should you agree, this should be the last change, promise;)

KSDaemon:
@oberstet I agree and just renamed it already :)

KSDaemon:
2 All: I renamed, removed, added, and edited different sections. Please have a look!

oberstet:
@KSDaemon this already looks quite good IMO! also, I really like the new name "Payload-Passthru"=) I still wanted to have a detailed read of the text in the PR, but I think we're "almost there". fwiw, I had a deep look into the respective code in Autobahn and Crossbar.io building on this feature, and I am certain we could match this spec quite easily.

Speaking of _using_ the "Payload-Passthru", e.g. for MQTT and for E2E (v2), I came up with the following precise attribute names and values mapping which we would do:

[wamp-ap-payload-passthru-schemes.pdf](https://github.com/wamp-proto/wamp-proto/files/9222459/wamp-ap-payload-passthru-schemes.pdf)

The idea is: the "Payload-Passthru" just defines the general working, plus provides a mapping table for predefined specific schemes. But those are then to be described in other sections.

Custom schemes can then reuse the basic machinery, as in: a router/client library might want to provide hooks for users to plugin their custom schemes.

Also, as you will note, the attributes are prefixed `ppt_` (for Payload-PassThru), because with MQTT, there isn't any encryption, but there is a special serializer "mqtt" which signals that the payload is the raw MQTT payload (only the payload, not the whole message). So I think this ppt prefix makes more sense.

What do you think?

KSDaemon:
@oberstet Looks really good! I agree with renaming. Now it reflects any possible use cases! I'll update the spec accordingly

oberstet:
one typo: the following messages have `Details|dict` not `Options|dict`:

* RESULT
* INVOCATION
* ERROR
* EVENT

KSDaemon:
> one typo: the following messages have `Details|dict` not `Options|dict`:
> 
> * RESULT
> * INVOCATION
> * ERROR
> * EVENT

My bad. Thnx! Fixed!

oberstet (approved):

oberstet:
@KSDaemon awesome, this is perfect! for me, this is ready to merge .. @ecorm what do you think?

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+
+`ppt_scheme` stands for Key Management Schema. It is required string attribute. This attribute can contain name or
+identifier of key management provider which is known to target peer, so it can be used to obtain information
+about encryption keys. *Router* understands that `Payload Passthru Mode` is in use
+by checking the existence and non-empty value of this attribute in `CALL`, `PUBLISH` and `YIELD` messages options.
+
+**ppt_serializer attribute**
+
+`ppt_serializer` attribute is optional. It specifies what serializer was used to build up payload object.
+It can be `mqtt`, `amqp`, `stomp` value just to inform that coming data is from related technologies or ordinary
```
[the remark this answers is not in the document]
ecorm: I don't think it's a good idea for the same attribute to be "overloaded" with two different meanings. Perhaps there should be a separate attribute for things such as `mqtt`, `amqp`, `stomp`, etc.

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+        ---
+        ERROR.Options.ppt_scheme|string
+        ERROR.Options.ppt_serializer|string
+        ERROR.Options.ppt_cipher|string
+        ERROR.Options.ppt_keyid|string
+
+
+**ppt_scheme attribute**
+
+`ppt_scheme` stands for Key Management Schema. It is required string attribute. This attribute can contain name or
```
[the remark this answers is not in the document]
ecorm: What is payload passthru mode is desired in a situation that doesn't involve encryption?

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+
+**ppt_serializer attribute**
+
+`ppt_serializer` attribute is optional. It specifies what serializer was used to build up payload object.
+It can be `mqtt`, `amqp`, `stomp` value just to inform that coming data is from related technologies or ordinary 
+`json`, `msgpack`, `cbor`, `flatbuffers` data serializers.
+
+**ppt_cipher attribute**
+
+`ppt_cipher` attribute is optional. It is required if payload is encrypted. This attribute specifies cryptographic
```
ecorm, line 132:
```suggestion
The `ppt_cipher` attribute is optional. It is required if the payload is encrypted. This attribute specifies the cryptographic
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+algorithm that was used to encrypt payload. It can be `xsalsa20poly1305`, `aes256gcm` for now.
```
ecorm, line 133:
```suggestion
algorithm that was used to encrypt the payload. It can be `xsalsa20poly1305`, `aes256gcm` for now.
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+
+**ppt_keyid attribute**
+
+`ppt_keyid` attribute is optional. This attribute can contain encryption key id that was used to encrypt payload.
```
ecorm, line 137:
```suggestion
The `ppt_keyid` attribute is optional. This attribute can contain the encryption key id that was used to encrypt the payload.
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+`ppt_keyid` attribute is always a type of string. The value can be a hex-encoded string, uri, DNS name,
```
ecorm, line 138:
```suggestion
The `ppt_keyid` attribute is a string type. The value can be a hex-encoded string, URI, DNS name, 
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+**ppt_cipher attribute**
+
+`ppt_cipher` attribute is optional. It is required if payload is encrypted. This attribute specifies cryptographic
+algorithm that was used to encrypt payload. It can be `xsalsa20poly1305`, `aes256gcm` for now.
+
+**ppt_keyid attribute**
+
+`ppt_keyid` attribute is optional. This attribute can contain encryption key id that was used to encrypt payload.
+`ppt_keyid` attribute is always a type of string. The value can be a hex-encoded string, uri, DNS name, 
+Ethereum address, UUID identifier - any meaningful value by which target peer can choose private key
```
ecorm, line 139:
```suggestion
Ethereum address, UUID identifier - any meaningful value which allows the target peer to a choose private key 
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+
+`ppt_cipher` attribute is optional. It is required if payload is encrypted. This attribute specifies cryptographic
+algorithm that was used to encrypt payload. It can be `xsalsa20poly1305`, `aes256gcm` for now.
+
+**ppt_keyid attribute**
+
+`ppt_keyid` attribute is optional. This attribute can contain encryption key id that was used to encrypt payload.
+`ppt_keyid` attribute is always a type of string. The value can be a hex-encoded string, uri, DNS name, 
+Ethereum address, UUID identifier - any meaningful value by which target peer can choose private key 
+without guessing. Format of the value may depend on `ppt_scheme` attribute.
```
ecorm, line 140:
```suggestion
without guessing. The format of the value may depend on the `ppt_scheme` attribute.
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+
+**ppt_ predefined schemes**
```
ecorm, line 142:
```suggestion
**ppt_ Predefined Schemes**
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+
+| Attribute      | Required? | Predefined Scheme (mqtt)  | Predefined Scheme (End-to-End Encryption) | Custom Scheme Example |
```
ecorm, line 144:
```suggestion
| Attribute      | Required? | Predefined Scheme (MQTT)  | Predefined Scheme (End-to-End Encryption) | Custom Scheme Example |
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+|----------------|-----------|---------------------------|-------------------------------------------|-----------------------|
+| ppt_scheme     | Y         | mqtt                      | wamp.eth                                  | x_my_ppt              |
```
ecorm, line 146:
```suggestion
| ppt_scheme     | Y         | MQTT                      | wamp.eth                                  | x_my_ppt              |
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+| ppt_serializer | N         | json, msgpack, cbor, mqtt | cbor, flatbuffers                         | custom                |
```
ecorm, line 147:
```suggestion
| ppt_serializer | N         | JSON, Msgpack, CBOR, MQTT | cbor, flatbuffers                         | custom                |
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+| ppt_cipher     | N         | -                         | xsalsa20poly1305, aes256gcm               | custom                |
+| ppt_keyid      | N         | -                         | *                                         | custom                |
+
+*: Least significant 20 bytes (160 bits) of SHA256 of public key (32 bytes) of data encryption key as HEX encoded
```
ecorm, line 151:
```suggestion
*: The least significant 20 bytes (160 bits) of the SHA256 of the public key (32 bytes) of the data encryption key, as a hex-encoded 
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+
+| Attribute      | Required? | Predefined Scheme (mqtt)  | Predefined Scheme (End-to-End Encryption) | Custom Scheme Example |
+|----------------|-----------|---------------------------|-------------------------------------------|-----------------------|
+| ppt_scheme     | Y         | mqtt                      | wamp.eth                                  | x_my_ppt              |
+| ppt_serializer | N         | json, msgpack, cbor, mqtt | cbor, flatbuffers                         | custom                |
+| ppt_cipher     | N         | -                         | xsalsa20poly1305, aes256gcm               | custom                |
+| ppt_keyid      | N         | -                         | *                                         | custom                |
+
+*: Least significant 20 bytes (160 bits) of SHA256 of public key (32 bytes) of data encryption key as HEX encoded 
+string with prefix `0x`, and upper/lower case letters encoding a checksum according to EIP55.
```
ecorm, line 152:
```suggestion
string with prefix `0x` and either uppercase/lowercase alphabetic characters, encoding a checksum according to EIP55.
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+
+**Message structure**
```
ecorm, line 154:
```suggestion
**Message Structure**
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+
+With `Payload Passthru Mode` in use message payload MUST BE sent as one binary item inside
```
ecorm, line 156:
```suggestion
When `Payload Passthru Mode` is in use, the message payload MUST be sent as one binary item within 
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+| ppt_cipher     | N         | -                         | xsalsa20poly1305, aes256gcm               | custom                |
+| ppt_keyid      | N         | -                         | *                                         | custom                |
+
+*: Least significant 20 bytes (160 bits) of SHA256 of public key (32 bytes) of data encryption key as HEX encoded 
+string with prefix `0x`, and upper/lower case letters encoding a checksum according to EIP55.
+
+**Message structure**
+
+With `Payload Passthru Mode` in use message payload MUST BE sent as one binary item inside 
+`Arguments|list`, while `ArgumentsKw|dict` MUST BE missing or empty.
```
ecorm, line 157:
```suggestion
`Arguments|list`, while `ArgumentsKw|dict` MUST be absent or empty.
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+            "ppt_keyid": "GTtQ37XGJO2O4R8Dvx4AUo8pe61D9evIWpKGQAPdOh0="
+        },
+        "com.myapp.secret_rpc_for_sensitive_data",
+        [Payload|binary]
+    ]
+```
+
+*Example.* Caller-to-Dealer progressive `CALL` with encryption and key ID.
+
+Nothing prevents to use `Payload Passthru Mode` with other features, for example `Progressive Calls`
```
ecorm, line 179:
```suggestion
Note that nothing prevents the use of `Payload Passthru Mode` with other features such as, for example, `Progressive Calls`
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+            "ppt_cipher": "xsalsa20poly1305",
+            "ppt_keyid": "GTtQ37XGJO2O4R8Dvx4AUo8pe61D9evIWpKGQAPdOh0=",
+            "progress": true
+        },
+        "com.myapp.progressive_rpc_for_sensitive_data",
+        [Payload|binary]
+    ]
+```
+
+*Example.* Caller-to-Dealer `CALL` with mqtt payload. Specifying `"ppt_serializer": "mqtt"` means, that
```
ecorm, line 198:
```suggestion
*Example.* Caller-to-Dealer `CALL` with MQTT payload. Specifying `"ppt_serializer": "mqtt"` means that 
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+            "ppt_keyid": "GTtQ37XGJO2O4R8Dvx4AUo8pe61D9evIWpKGQAPdOh0=",
+            "progress": true
+        },
+        "com.myapp.progressive_rpc_for_sensitive_data",
+        [Payload|binary]
+    ]
+```
+
+*Example.* Caller-to-Dealer `CALL` with mqtt payload. Specifying `"ppt_serializer": "mqtt"` means, that 
+original mqtt message payload is passed as WAMP payload message without any transcoding, as is.
```
ecorm, line 199:
```suggestion
the original mqtt message payload is passed as a WAMP payload message as is, without any transcoding.
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+        {
+            "ppt_scheme": "mqtt",
+            "ppt_serializer": "mqtt"
+        },
+        "com.myapp.mqtt_processing",
+        [Payload|binary]
+    ]
+```
+
+*Example.* Caller-to-Dealer `CALL` with mqtt payload. Specifying `"ppt_scheme": "mqtt"` simply indicates, that
```
ecorm, line 215:
```suggestion
*Example.* Caller-to-Dealer `CALL` with MQTT payload. Specifying `"ppt_scheme": "mqtt"` simply indicates that
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+original source of payload data is received from related system. Specifying `"ppt_serializer": "json"` means, that
```
ecorm, line 216:
```suggestion
the original source of payload data is received from a related system. Specifying `"ppt_serializer": "json"` means that
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+original mqtt message payload was parsed and encoded with `json` serializer.
```
ecorm, line 217:
```suggestion
the original MQTT message payload was parsed and encoded with the `json` serializer.
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+            "ppt_scheme": "wamp.eth",
+            "ppt_serializer": "cbor",
+            "ppt_cipher": "xsalsa20poly1305",
+            "ppt_keyid": "GTtQ37XGJO2O4R8Dvx4AUo8pe61D9evIWpKGQAPdOh0="
+        },
+        [Payload|binary]
+    ]
+```
+
+*Example.* Dealer-to-Callee `INVOCATION` with mqtt payload
```
ecorm, line 251:
```suggestion
*Example.* Dealer-to-Callee `INVOCATION` with MQTT payload
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+            "ppt_cipher": "xsalsa20poly1305",
+            "ppt_keyid": "GTtQ37XGJO2O4R8Dvx4AUo8pe61D9evNSpGMDQWdOh1="
+        },
+        [Payload|binary]
+    ]
+```
+
+*Example.* Callee-to-Dealer progressive `YIELD` with encryption and key ID
+
+Nothing prevents to use `Payload Passthru Mode` with other features, for example `Progressive Call Results`
```
ecorm, line 286:
```suggestion
Nothing prevents the use of `Payload Passthru Mode` with other features such as, for example, `Progressive Call Results`
```

ecorm: This is sentence is repeated multiple times through the examples. It should be consolidated in one place.

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+            "ppt_cipher": "xsalsa20poly1305",
+            "ppt_keyid": "GTtQ37XGJO2O4R8Dvx4AUo8pe61D9evNSpGMDQWdOh1="
+        },
+        [Payload|binary]
+    ]
+```
+
+*Example.* Dealer-to-Caller progressive `RESULT` with encryption and key ID
+
+Nothing prevents to use `Payload Passthru Mode` with other features, for example `Progressive Call Results`
```
ecorm, line 323:
```suggestion
Nothing prevents the use of `Payload Passthru Mode` with other features such as, for example, `Progressive Call Results`.
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+            "ppt_scheme": "wamp.eth",
+            "ppt_serializer": "flatbuffers",
+            "ppt_cipher": "xsalsa20poly1305",
+            "ppt_keyid": "GTtQ37XGJO2O4R8Dvx4AUo8pe61D9evNSpGMDQWdOh1="
+        },
+        [Payload|binary]
+    ]
+```
+
+**About supported serializers and cryptographic ciphers**
```
ecorm, line 396:
```suggestion
**About Supported Serializers and Cryptographic Ciphers**
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+
+WAMP works as infrastructure for delivering messages between peers. Regardless of what encryption algorithm
```
ecorm, line 398:
```suggestion
WAMP serves as infrastructure for delivering messages between peers. Regardless of what encryption algorithm 
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+            "ppt_keyid": "GTtQ37XGJO2O4R8Dvx4AUo8pe61D9evNSpGMDQWdOh1="
+        },
+        [Payload|binary]
+    ]
+```
+
+**About supported serializers and cryptographic ciphers**
+
+WAMP works as infrastructure for delivering messages between peers. Regardless of what encryption algorithm 
+and serializer were chosen for `Payload Passthru Mode`, *Router* will not inspect and analyze related encryption
```
ecorm, line 399:
```suggestion
and serializer were chosen for `Payload Passthru Mode`, a *Router* shall not inspect and analyze the `ppt_` options and payload
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+        },
+        [Payload|binary]
+    ]
+```
+
+**About supported serializers and cryptographic ciphers**
+
+WAMP works as infrastructure for delivering messages between peers. Regardless of what encryption algorithm 
+and serializer were chosen for `Payload Passthru Mode`, *Router* will not inspect and analyze related encryption 
+message options and payload. It is up to an application side code responsibility to use serializers and
```
ecorm, line 400:
```suggestion
of encypted messages. The application is responsible for choosing serializers and 
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,401 @@
...
+        [Payload|binary]
+    ]
+```
+
+**About supported serializers and cryptographic ciphers**
+
+WAMP works as infrastructure for delivering messages between peers. Regardless of what encryption algorithm 
+and serializer were chosen for `Payload Passthru Mode`, *Router* will not inspect and analyze related encryption 
+message options and payload. It is up to an application side code responsibility to use serializers and 
+ciphers known to every peer involved into message processing.
```
ecorm, line 401:
```suggestion
ciphers known to every peer involved in message processing.
```

ecorm (changes requested):
Grammar fixes and readability suggestions.

ecorm:
> @ecorm what do you think

Made a pass to fix grammar and improve readability, with about a million missing "a" or "the" articles :wink: (yes I understand they're not used in Slavic languages, just poking fun).

I have two technical critiques:
- The `ppt_scheme` attribute is required to be non-empty, and its presence is what "triggers" Payload Passthru Mode. What if an application wants to enable Payload Passthru Mode, but doesn't want/need to use encryption?
- I don't like that `ppt_serializer` is "overloaded" for two different meanings: identifying the message transport technology (MQTT, etc) or identifying the actual serializer. Could it be split into two attributes?

rfc/text/advanced/ap_50_payload_passthru_mode.md
KSDaemon: I think about skipping this, as it shows a value that we use. And under a hidden agreement the values are lowercase

rfc/text/advanced/ap_50_payload_passthru_mode.md
KSDaemon: The same here: we specify values in lowercase

rfc/text/advanced/ap_50_payload_passthru_mode.md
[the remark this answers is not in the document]
KSDaemon: Well, maybe... But i think it is a rare case anyway... 
We can think and discuss it...

rfc/text/advanced/ap_50_payload_passthru_mode.md
ecorm: Sorry, I thought you were naming those technologies. I see now that they are identifiers, so yes lowercase.

ecorm:
Another technical critique is that there needs to be more clarification about what to do when the WAMP-level messages are serialized as JSON. I'm guessing the payload should be Base64-encoded, but this needs to be stated somewhere explicitly (unless I missed it).

Another corner case that needs clarification is when WAMP-level messages are JSON, and passthru payloads are also JSON.

KSDaemon:
Phew!! I've done it!)))

KSDaemon:
@ecorm 
> The ppt_scheme attribute is required to be non-empty, and its presence is what "triggers" Payload Passthru Mode. What if an application wants to enable Payload Passthru Mode, but doesn't want/need to use encryption?
In that case app should just specify some value there. Maybe we should explicitly add `passthru` value for such cases? 
@oberstet What do you think?

> I don't like that ppt_serializer is "overloaded" for two different meanings: identifying the message tunneling technology (MQTT, etc) or identifying the actual serializer. Could it be split into two attributes?

For identifying tech there is a `ppt_scheme` attribute. So serializer attribute is only for specifying serializers in my opinion.

KSDaemon:
> Another technical critique is that there needs to be more clarification about what to do when the WAMP-level messages are serialized as JSON. I'm guessing the payload should be Base64-encoded (with nul character prefix), but this needs to be stated somewhere explicitly (unless I missed it).
> 
> Another corner case that needs clarification is when WAMP-level messages are JSON, and passthru payloads are also JSON.

These are  great questions!

KSDaemon:
Tried to find some words in spec about encoding binary data into JSON... No luck...
So.. Yeah, I agree, let's explicitly say about base64 encoding of binary data when JSON serializer is used in WAMP session.

Regarding JSON-in-JSON — we can state that payload may be string or binary or we can specify the requirement for prepending null character for example to JSON-string...

oberstet:
>     * The `ppt_scheme` attribute is required to be non-empty, and its presence is what "triggers" Payload Passthru Mode. What if an application wants to enable Payload Passthru Mode, but doesn't want/need to use encryption?

then the application can use `ppt_scheme` without using `ppt_cipher` and `ppt_keyid`

>     * I don't like that `ppt_serializer` is "overloaded" for two different meanings: identifying the message tunneling technology (MQTT, etc) or identifying the actual serializer. Could it be split into two attributes?

ok, I can follow that argument. however, in this case, how about the following? which doesn't intro a new attribute:

```
ppt_scheme == "mqtt", ppt_serializer == "native"
```

for specifying MQTT, with a "native" payload, that is native and custom to the application ("native" in as: the WAMP client library will also not touch it .. deliver raw bytes to/from the app code, rather than applying the specified `ppt_serializer` before delivering to/from app code)

-----

> Another technical critique is that there needs to be more clarification about what to do when the WAMP-level messages are serialized as JSON.

so when `ppt_serializer` is specified `!= "mqtt/native"`, eg `"json"`, then the serializer is applied first (by the WAMP client library), and then (without encryption), the resulting binary payload is embedded in a WAMP message, which might use a different serializer.

if that 2nd WAMP message level serializer is JSON, then the implementation must support

https://github.com/wamp-proto/wamp-proto/blob/master/rfc/text/advanced/ap_41_serialization_binary_json.md

however, for some reasons, it seems this file isn't currently referenced in the spec top-level texts, and hence not rendered. which is a bug;)

oberstet:
> Yeah, I agree, let's explicitly say about base64 encoding of binary data when JSON serializer is used in WAMP session.

the "Binary support in JSON" section has been currently lost in the spec (pls see above). this is a bug.

however, that still doesn't answer the question of how the binary `payload` resulting from applying `ppt_serializer` is to be treated in `args[0]` in the embedded WAMP message - treated wrt the serializer used for the WAMP message

I would argue: the embedded WAMP message serializer MUST support binary payloads, and for JSON, that means, the implementation MUST support the "Binary support in JSON" feature

I would not want to pay for Base64 encoding of _all_ my WAMP-JSON traffic by _deactivating_ this feature. I mean, yeah, I would use CBOR anyways, but ... - at least this is my argument.

The problem arises necessarily since: the type of `payload` is always binary, but the outer embedded WAMP message serializer might not be able to transport that unencoded

KSDaemon:
@oberstet @ecorm I added more text covering what we discussed above. Yep, that is not completed, but please have a look and maybe provide some feedback or additions to this. Let's get together and make one last push!

KSDaemon:
@ecorm Your thoughts on the last updates?

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,422 @@
...
+**ppt_keyid Attribute**
+
+The `ppt_keyid` attribute is optional. This attribute can contain the encryption key id that was used to encrypt 
+the payload. The `ppt_keyid` attribute is a string type. The value can be a hex-encoded string, URI, DNS name, 
+Ethereum address, UUID identifier - any meaningful value which allows the target peer to choose a private key 
+without guessing. The format of the value may depend on the `ppt_scheme` attribute.
+
+**ppt_ Predefined Schemes**
+
+| Attribute      | Required? | Predefined Scheme (MQTT)    | Predefined Scheme (End-to-End Encryption) | Custom Scheme Example |
```
ecorm, line 147: It's not clear if MQTT and EtoE are just examples or are mandated by the WAMP spec.

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,422 @@
...
+| ppt_scheme     | Y         | mqtt                        | wamp.eth                                  | x_my_ppt              |
+| ppt_serializer | N         | json, msgpack, cbor, native | cbor, flatbuffers                         | custom                |
+| ppt_cipher     | N         | -                           | xsalsa20poly1305, aes256gcm               | custom                |
+| ppt_keyid      | N         | -                           | *                                         | custom                |
+
+*: The least significant 20 bytes (160 bits) of the SHA256 of the public key (32 bytes) of the data encryption key, 
+as a hex-encoded string with prefix `0x` and either uppercase/lowercase alphabetic characters, encoding a 
+checksum according to EIP55.
+
+When `Payload Passthru Mode` is used for gateways to other technologies like MQTT Brokers or AMQP Queues and
```
ecorm, line 158:
```suggestion
When `Payload Passthru Mode` is used for gateways to other technologies, such as MQTT Brokers or AMQP Queues, then
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,422 @@
...
+the `ppt_serializer` attribute may be set to `native` value. This means that payload is not touched by WAMP peers,
```
ecorm, line 159:
```suggestion
the `ppt_serializer` attribute may be set to the `native` value. This means that the payload is not to be modified by WAMP peers, 
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,422 @@
...
+| ppt_cipher     | N         | -                           | xsalsa20poly1305, aes256gcm               | custom                |
+| ppt_keyid      | N         | -                           | *                                         | custom                |
+
+*: The least significant 20 bytes (160 bits) of the SHA256 of the public key (32 bytes) of the data encryption key, 
+as a hex-encoded string with prefix `0x` and either uppercase/lowercase alphabetic characters, encoding a 
+checksum according to EIP55.
+
+When `Payload Passthru Mode` is used for gateways to other technologies like MQTT Brokers or AMQP Queues and
+the `ppt_serializer` attribute may be set to `native` value. This means that payload is not touched by WAMP peers, 
+not serialized in any manner and is delivered as is from the originator peer. Another possible case is when
```
ecorm, line 160:
```suggestion
nor serialized in any manner, and is delivered as-is from the originating peer. Another possible case is when
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,422 @@
...
+the `ppt_serializer` attribute is set to any valid serializer, for example `msgpack`. In this case WAMP originator
```
ecorm, line 161:
```suggestion
the `ppt_serializer` attribute is set to any valid serializer, for example `msgpack`. In this case, the originating WAMP
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,422 @@
...
+client peer first applies `ppt_serializer` to serialize payload (without encryption), then the resulting
```
ecorm, line 162:
```suggestion
client peer first applies `ppt_serializer` to serialize the payload (without encryption), then the resulting
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,422 @@
...
+binary payload is embedded in a WAMP message, which might use different serializer that was chosen during
```
ecorm, line 163:
```suggestion
binary payload is embedded in the WAMP message, the latter having possibly a different serializer depending on the one chosen during
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,422 @@
...
+WAMP Session establishment.
+
+**Important note regarding JSON serializer**
+
+With `Payload Passthru Mode` payloads are treated as binaries. To send binaries WAMP session serializer MUST
```
ecorm, line 168:
```suggestion
With `Payload Passthru Mode`, payloads are treated as binary. To send these binary payloads, the WAMP session serializer MUST
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,422 @@
...
+support binary types. Most of the serializers support binary types, but JSON serializer does not. To use
```
ecorm, line 169:
```suggestion
support byte arrays. Most serialization formats known to WAMP support byte arrays, but JSON does not support them natively. To use
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,422 @@
...
+`Payload Passthru Mode` with JSON serializer WAMP peers MUST support
```
ecorm, line 170:
```suggestion
`Payload Passthru Mode` with a JSON serializer, WAMP peers MUST perform the special
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,422 @@
...
+client peer first applies `ppt_serializer` to serialize payload (without encryption), then the resulting
+binary payload is embedded in a WAMP message, which might use different serializer that was chosen during
+WAMP Session establishment.
+
+**Important note regarding JSON serializer**
+
+With `Payload Passthru Mode` payloads are treated as binaries. To send binaries WAMP session serializer MUST
+support binary types. Most of the serializers support binary types, but JSON serializer does not. To use
+`Payload Passthru Mode` with JSON serializer WAMP peers MUST support 
+[Binary serialization in JSON](#binary-support-in-json). This come with huge overhead, so general advice is
```
ecorm, line 171:
```suggestion
[Binary serialization in JSON](#binary-support-in-json). This conversion may have unacceptable overhead, so it is generally advised
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,422 @@
...
+to use WAMP session serializers with native binary support, for example `MessagePack`, `CBOR`, `FlatBuffers`.
```
ecorm, line 172:
```suggestion
to use WAMP session serializers with native byte array support, for example, `MessagePack`, `CBOR`, or `FlatBuffers`.
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -0,0 +1,422 @@
...
+
+When `Payload Passthru Mode` is used for gateways to other technologies like MQTT Brokers or AMQP Queues and
+the `ppt_serializer` attribute may be set to `native` value. This means that payload is not touched by WAMP peers, 
+not serialized in any manner and is delivered as is from the originator peer. Another possible case is when
+the `ppt_serializer` attribute is set to any valid serializer, for example `msgpack`. In this case WAMP originator
+client peer first applies `ppt_serializer` to serialize payload (without encryption), then the resulting
+binary payload is embedded in a WAMP message, which might use different serializer that was chosen during
+WAMP Session establishment.
+
+**Important note regarding JSON serializer**
```
ecorm, line 166:
```suggestion
**Important Note Regarding JSON Serialization**
```

rfc/text/advanced/ap_50_payload_passthru_mode.md
ecorm: This would be best split into separate tables to fit the page width in RFC text format.

ecorm (changes requested):
Grammar and readability of latest changes concerning serializers.

KSDaemon:
Okay! Fixed that! Hope it was the last round %)

KSDaemon:
OMG! It is merged! 🙌🏻

PR #419, opened 2022-08-03
add e2e encryption section [WIP]

oberstet:


oberstet:
It seems `rfc/text/advanced/ap_51_e2e_encryption.md` collides with

https://github.com/wamp-proto/wamp-proto/pull/417/files#diff-50bfe4d488b02505c2966e76581fdf7854ee87004b751869f488f4fbfafa3211

@KSDaemon could you maybe remove the (empty) file above from the PPT PR?

KSDaemon:
> maybe I just wait a little bit, so we have the PPT PR hopefully merged then already ...

Yeah! I think it is the best choice!) Hope we'll land that PR soon!

oberstet:
added a figure that describes how keys etc are related .. this will be fun to transform to ASCII ;)

https://github.com/wamp-proto/wamp-proto/blob/c964c89a9d0ffa7d8da43f4ef04dd0f56a1a922e/_work/ap_51_e2e_encryption.pdf

besides this figure, I'm planning for a 2nd: illustrating the 2 cases of trustroots: standalone and on-chain

oberstet:
TODO from feedback via discord - Part 1:

* [ ] redraw  Fig. 1 - Fig. 3 from **ap_51_e2e_encryption.odg** in ASCII using https://asciiflow.com/
* [ ] Fig. 1: add text describing numbered items in **ap_51_e2e_encryption.odg**
* [ ] Fig. 2: add text:
- state what are Client CS keys, 
- what are Delegate ETH keys, 
- why there is two pairs, 
- description of why ETH was chosen for Peer identification 
- and some general overview of how certificate chain works.
* [ ] Fig. 3: add text:
- general description of standalone trustroot and shared. May be provide typical use cases for them
- what is User and what is Delegate
- scheme contains new terms for a first time (in yellow baloon): types of trust establishment. So we need to describe it, maybe just give a short description to each and refer user to next sections for details.

oberstet:
TODO from feedback via discord - Part 2:

* [ ] redraw  Fig. 4 - Fig. 6 (Scenario 0, 1 and 2) from **ap_51_e2e_encryption.odg** in ASCII using https://asciiflow.com/
* [ ] Fig. 4: add text:
- What does dotted blue arrow mean?
- There are 3 R2R links, but only one of them is with both directions. Why links from Edge devices are one-way? If you mean to show the direction of connections, then it confuses a little bit with R2R links, because via one link messages can flow in any direction
- Left WAMP Client (grey): if Edge Device itself is a WAMP client, then it connects to Cloud Server via regular WAMP Link, not the R2R Link.
- We see User 1. But how does he relate to all other items on the scheme? Now he is just aside of all.
- There is a mention of Root CA with Centralized Trust, but only in context of User 1 and no relation to other scheme parts. Maybe it will be great to add another user 2 associated with e.g. WAMP Client with his own Certificate and show that both certificates of users are singed and trusted via common Root CA
- Yellow baloon for upper right Cloud WAMP Node: why doesn't left Cloud Node have same one?
- Yellow baloon for right Edge Server Node: What does it mean “runs on-demand only”? And what does it specificaly mean “E2EE” there, if there is a whole main ellipse stating for E2EE App Realm in the center of scheme? Maybe it (E2EE text) just need to be removed from there. 
* [ ] Fig. 5: add text:
- mostly the same questions as in previous one + next:
- We see User 2 and User 3, but I suggest to distinguish WAMP Regular Client User and WAMP Router Provider. Maybe we should rename User 2 and User 3 to WAMP Relay Server Provider 1 and 2. So it will be clear: there are consumers and providers.
- I also suggest to provide summary for “Private Realm”. What it means? Basic words about its visibility, access control and all that stuff. The same applies to all scenarios.
- Also we should describe what it means “On-chain user”. And provide information about “Realm owner role”
* [ ] Fig. 6: add text:
- mostly the same questions as in previous two + next:
- Need to describe roles of all Users.

oberstet:
rgd certificates and trustroots, which are used in above E2EE proposal, as well as used in a new WAMP-Cryptosign feature

https://github.com/wamp-proto/wamp-proto/blob/master/rfc/text/advanced/ap_35_authentication_cryptosign.md#trustroots-and-certificates-trustrootcerts

the master branches of AutobahnPython and Crossbar.io now contain working code - here is a complete example

https://github.com/crossbario/crossbar-examples/tree/master/authentication/cryptosign/trustroot

on the router side, the only thing configured is a self-signed certificate as the trustroot for a realm

https://github.com/crossbario/crossbar-examples/blob/3dbda8d44f4c8ac261714a97fd94f6d6462743c1/authentication/cryptosign/trustroot/.crossbar/config-standalone-trustroot.json#L55

a client can then authenticate using WAMP-Cryptosign, without the router knowing anything about the client before, just because the client presented a complete certificate chain, ending in the same trustroot, and starting from a delegate certificate that covers the clients WAMP-Cryptosign public key (which is unknown previously to the router)

https://github.com/crossbario/crossbar-examples/blob/3dbda8d44f4c8ac261714a97fd94f6d6462743c1/authentication/cryptosign/trustroot/client.py#L85

oberstet:
@KSDaemon awesome!! the diagram looks really great, and quite a bit of work I would think. thanks a lot=) I've looked through all the changes you've added to this PR, just 2 minor comments

```
-ppt_scheme      = "wamp.eth"
+ppt_scheme      = "wamp"
```

so we go with "wamp" for ppt_scheme, and drop the "eth" suffix? I'm good with that, just asking, and I forgot our final conclusion in the discussion of the topic we had ..

> *Overview of End-to-End Encryption flow*

yes. that's about right .. roughly. as in: there are different options e.g. when/how the peer creating new symmetric payload encryption keys decides to do so. It might rotate a single key used for any results or events every hour. It might rotate a key for each different caller of a given procedure every 10 calls. Many more options. but this can be laid out in additional/later paragraphs ..

KSDaemon:
Cool! 
@oberstet Well, we agreed to go with `ppt_scheme = "wamp"`. That's why i changed it.

> yes. that's about right .. roughly. as in: there are different options e.g. when/how the peer creating new symmetric payload encryption keys decides to do so. It might rotate a single key used for any results or events every hour. It might rotate a key for each different caller of a given procedure every 10 calls. Many more options. but this can be laid out in additional/later paragraphs ..

Exactly. As a new developer comes to our spec and start reading it, it would be cool to show the overall design of E2EE and describe the main flow providing points that can be done in different manner and developer can (if he/she wants to) read about that later.

oberstet:
> show the overall design of E2EE and describe the main flow providing points that can be done in different manner

yes! absoutely, I agree that's what we should target. the approach of providing elements, which can be freely combined in an orthogonal fashion, is really a strength of WAMP / the spec in my eyes.

and one of those elements in E2EE is policy/schedule of key rotation performed by responders (callees/publishers)

KSDaemon:
Hey @oberstet whatsup?) Moving this elsewhere?

PR #423, opened 2022-09-16
Fixed ppt examples schemas and messages

KSDaemon:
I've started implementing ppt feature and faced a few inaccuracies in spec. 

Fixed some misconceptions and added more description to ppt mode example schemas.

### DIFF
--- a/rfc/text/advanced/ap_50_payload_passthru_mode.md
+++ b/rfc/text/advanced/ap_50_payload_passthru_mode.md
@@ -207,6 +207,22 @@ generally advised to use WAMP session serializers with native byte array support
 When `Payload Passthru Mode` is in use, the message payload MUST be sent as one binary item within 
 `Arguments|list`, while `ArgumentsKw|dict` MUST be absent or empty.
 
+Since many WAMP messages assume the possibility of simultaneous use of `Arguments|list` and `ArgumentsKw|dict`,
+WAMP client implementations must package arguments into the following hash table and then serialize it and
+transmit as a single element within `Arguments|list`.
+
+{align="left"}
+```json
+{
+    "args": Arguments|list,
+    "kwargs": ArgumentsKw|dict
+}
+```
+
+This will allow maintaining a single interface for client applications, regardless of whether the
+```Payload Passthru Mode``` mode, or especially `Payload End-to-End Encryption` which is built on top of
+`Payload End-to-End Encryption` is used or not.
+
 *Example.* Caller-to-Dealer `CALL` with encryption and key ID
 
 {align="left"}
--- a/rfc/text/advanced/ap_50_payload_passthru_mode.md
+++ b/rfc/text/advanced/ap_50_payload_passthru_mode.md
@@ -121,15 +121,17 @@ as well as `ERROR` messages in the case of failures.
 The `ppt_scheme` identifies the Payload Schema. It is a required string attribute. For End-2-End Encryption flow
 this attribute can contain the name or identifier of a key management provider that is known to the target peer, 
 so it can be used to obtain information about encryption keys. For gateways and external schemas this can
-contain the name of related technology such as `mqtt`, `amqp`, `stomp`. A *Router* can recognize that 
-`Payload Passthru Mode` is in use by checking the existence and non-empty value of this attribute within the 
-options of `CALL`, `PUBLISH` and `YIELD` messages.
+contain the name of related technology. The one predefined is `mqtt`. Others may be introduced later. 
+A *Router* can recognize that `Payload Passthru Mode` is in use by checking the existence and non-empty value of 
+this attribute within the options of `CALL`, `PUBLISH` and `YIELD` messages.
 
 **ppt_serializer Attribute**
 
 The `ppt_serializer` attribute is optional. It specifies what serializer was used to encode the payload.
 It can be a `native` value to indicate that the incoming data is tunneling through other technologies
 specified by the `ppt_scheme`, or it can be ordinary `json`, `msgpack`, `cbor`, `flatbuffers` data serializers.
+For some predefined `ppt_scheme` schemas this option may be omitted as schema defines the concrete serializer. 
+See predefined schemas below.
 
 **ppt_cipher Attribute**
 
@@ -149,15 +151,12 @@ without guessing. The format of the value may depend on the `ppt_scheme` attribu
 
 {align="left"}
 
-| Attribute      | Required? | Value                        |
-|----------------|-----------|------------------------------|
-| ppt_scheme     | Y         | mqtt                         |
-| ppt_serializer | N*        | json, msgpack, cbor, native* |
-| ppt_cipher     | N         | -                            |
-| ppt_keyid      | N         | -                            |
-
-*: If `ppt_serializer` is not provided then it is assuming as `native`. So no additional serialization 
-will be applied to payload and payload will be serialized within WAMP message with session serializer.
+| Attribute      | Required? | Value  |
+|----------------|-----------|--------|
+| ppt_scheme     | Y         | mqtt   |
+| ppt_serializer | N*        | native |
+| ppt_cipher     | N         | -      |
+| ppt_keyid      | N         | -      |
 
 **End-to-End Encryption Predefined Scheme**
 
@@ -166,14 +165,11 @@ will be applied to payload and payload will be serialized within WAMP message wi
 | Attribute      | Required? | Value                       |
 |----------------|-----------|-----------------------------|
 | ppt_scheme     | Y         | wamp.eth                    |
-| ppt_serializer | N         | cbor, flatbuffers, native*  |
+| ppt_serializer | Y         | cbor, flatbuffers           |
 | ppt_cipher     | N         | xsalsa20poly1305, aes256gcm |
-| ppt_keyid      | N         | **                          |
+| ppt_keyid      | N         | *                           |
 
-*: If `ppt_serializer` is not provided then it is assuming as `native`. So payload will be serialized 
-with session serializer, then encrypted and sent as binary in WAMP message serialized by session serializer.
-
-**: The least significant 20 bytes (160 bits) of the SHA256 of the public key (32 bytes) of the data encryption key,
+*: The least significant 20 bytes (160 bits) of the SHA256 of the public key (32 bytes) of the data encryption key,
 as a hex-encoded string with prefix `0x` and either uppercase/lowercase alphabetic characters, encoding a
 checksum according to EIP55.
 
@@ -189,7 +185,7 @@ checksum according to EIP55.
 | ppt_keyid      | N         | custom   |
 
 
-When `Payload Passthru Mode` is used for gateways to other technologies, such as MQTT Brokers or AMQP Queues, then
+When `Payload Passthru Mode` is used for gateways to other technologies, such as MQTT Brokers, then
 the `ppt_serializer` attribute may be set to the `native` value. This means that the payload is not to be modified 
 by WAMP peers, nor serialized in any manner, and is delivered as-is from the originating peer. Another possible case 
 is when the `ppt_serializer` attribute is set to any valid serializer, for example `msgpack`. In this case the 
--- a/rfc/text/advanced/ap_50_payload_passthru_mode.md
+++ b/rfc/text/advanced/ap_50_payload_passthru_mode.md
@@ -154,7 +154,7 @@ without guessing. The format of the value may depend on the `ppt_scheme` attribu
 | Attribute      | Required? | Value  |
 |----------------|-----------|--------|
 | ppt_scheme     | Y         | mqtt   |
-| ppt_serializer | N*        | native |
+| ppt_serializer | N         | native |
 | ppt_cipher     | N         | -      |
 | ppt_keyid      | N         | -      |
 
--- a/rfc/text/advanced/ap_50_payload_passthru_mode.md
+++ b/rfc/text/advanced/ap_50_payload_passthru_mode.md
@@ -151,12 +151,15 @@ without guessing. The format of the value may depend on the `ppt_scheme` attribu
 
 {align="left"}
 
-| Attribute      | Required? | Value  |
-|----------------|-----------|--------|
-| ppt_scheme     | Y         | mqtt   |
-| ppt_serializer | N         | native |
-| ppt_cipher     | N         | -      |
-| ppt_keyid      | N         | -      |
+| Attribute      | Required? | Value                        |
+|----------------|-----------|------------------------------|
+| ppt_scheme     | Y         | mqtt                         |
+| ppt_serializer | N*        | native, json, msgpack, cbor  |
+| ppt_cipher     | N         | -                            |
+| ppt_keyid      | N         | -                            |
+
+*: If `ppt_serializer` is not provided then it is assuming as `native`. So no additional serialization
+will be applied to payload and payload will be serialized within WAMP message with session serializer.
 
 **End-to-End Encryption Predefined Scheme**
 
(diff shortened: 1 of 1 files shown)


KSDaemon:
@ecorm Will be glad if you also can have a look as you already know the context ;)

ecorm:
@KSDaemon I'll see if I have some moments to spare this weekend for review.

I removed myself from he org because I am not comfortable with this DAO business and the movement towards cyptocurrency. I was never asked if I wanted to join the org in the first place. I prefer to remain as a contributor.

I'm also considering writing my own standard inspired by WAMP, but without the complexity of routed RPCs (insert obligatory XKCD comic about yet another standard). I don't want there to be a conflict of interest, so it's best that I just remain as contributor with no merging privileges.

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -118,17 +118,18 @@ as well as `ERROR` messages in the case of failures.
...
 
-The `ppt_scheme` identifies the Key Management Schema. It is a required string attribute. This attribute can contain 
-the name or identifier of a key management provider that is known to the target peer, so it can be used to obtain 
-information about encryption keys. A *Router* can recognize that `Payload Passthru Mode` is in use
-by checking the existence and non-empty value of this attribute within the options of `CALL`, `PUBLISH` and `YIELD` 
-messages.
+The `ppt_scheme` identifies the Payload Schema. It is a required string attribute. For End-2-End Encryption flow
+this attribute can contain the name or identifier of a key management provider that is known to the target peer, 
+so it can be used to obtain information about encryption keys. For gateways and external schemas this can
+contain the name of related technology such as `mqtt`, `amqp`, `stomp`. A *Router* can recognize that
```
oberstet, line 124: I would only refer to mqtt, not the other ones, and only a mapping for mqtt is defined, and not for the other ones (and there is no interest anyways I guess in those)

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -148,12 +149,15 @@ without guessing. The format of the value may depend on the `ppt_scheme` attribu
...
-| ppt_cipher     | N         | -                           |
-| ppt_keyid      | N         | -                           |
+| Attribute      | Required? | Value                        |
+|----------------|-----------|------------------------------|
+| ppt_scheme     | Y         | mqtt                         |
+| ppt_serializer | N*        | json, msgpack, cbor, native* |
+| ppt_cipher     | N         | -                            |
+| ppt_keyid      | N         | -                            |
+
+*: If `ppt_serializer` is not provided then it is assuming as `native`. So no additional serialization
```
oberstet, line 159: the possible values is already defined by the ppt_schemes "wamp.eth" and "mqtt". thus, the don't need to specify what happens if it is "not provided", since it depends on ppt_scheme

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -162,11 +166,14 @@ without guessing. The format of the value may depend on the `ppt_scheme` attribu
 | Attribute      | Required? | Value                       |
 |----------------|-----------|-----------------------------|
 | ppt_scheme     | Y         | wamp.eth                    |
-| ppt_serializer | N         | cbor, flatbuffers           |
+| ppt_serializer | N         | cbor, flatbuffers, native*  |
```
oberstet, line 169: wamp.eth doesn't allow "native" for ppt_serializer

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -162,11 +166,14 @@ without guessing. The format of the value may depend on the `ppt_scheme` attribu
...
 | ppt_cipher     | N         | xsalsa20poly1305, aes256gcm |
-| ppt_keyid      | N         | *                           |
+| ppt_keyid      | N         | **                          |
+
+*: If `ppt_serializer` is not provided then it is assuming as `native`. So payload will be serialized
```
oberstet, line 173: this is a new meaning for "native": use the same serializer as in _the_ enclosing WAMP session - which doesn't make sense, as there are at least 2 such sessions

"native" means: defined outside of WAMP by the ppt_scheme in use, and the specific scheme "wamp.eth" only allows "cbor" and "flatbuffers" - not "native"

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -244,8 +251,8 @@ for example, `Progressive Calls`.
     ]
 ```
 
-*Example.* Caller-to-Dealer `CALL` with MQTT payload. Specifying `"ppt_serializer": "mqtt"` means that 
-the original MQTT message payload is passed as WAMP payload message as is, without any transcoding, as is.
+*Example.* Caller-to-Dealer `CALL` with MQTT payload. Specifying `"ppt_serializer": "native"` means that 
+the original MQTT message payload is passed as WAMP payload message as is, without any transcoding.
```
oberstet, line 255: exactly! thus, "native" means "whatever the app/tech outside wamp defines", such as in case of mqtt is "whatever the app defines", since mqtt is leaving that fully to the app

oberstet (changes requested):

oberstet:
> I removed myself from he org because I am not comfortable with this DAO business

ok, cool! just for the record: you are on the list of contributors in the spec text, you had been member of the github admin team here, but you never had been member of the DAO.

oberstet:
rgd ppt_serializer, the idea is to define a couple of useful values which then can be used in ppt_schemes, but the ppt_scheme might then further restrict the use of those predefined values, eg

* wamp.eth: https://wamp-proto.org/wamp_latest_ietf.html#section-14.1-34
* mqtt: https://wamp-proto.org/wamp_latest_ietf.html#section-14.1-32

in above, only the ppt_scheme "mqtt" uses the ppt_serializer "native", which then means: whatever the app uses (because mqtt itself also doesn't say anything specific about the payload)

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -118,17 +118,18 @@ as well as `ERROR` messages in the case of failures.
...
 **ppt_scheme Attribute**
 
-The `ppt_scheme` identifies the Key Management Schema. It is a required string attribute. This attribute can contain 
-the name or identifier of a key management provider that is known to the target peer, so it can be used to obtain 
-information about encryption keys. A *Router* can recognize that `Payload Passthru Mode` is in use
-by checking the existence and non-empty value of this attribute within the options of `CALL`, `PUBLISH` and `YIELD` 
-messages.
+The `ppt_scheme` identifies the Payload Schema. It is a required string attribute. For End-2-End Encryption flow
+this attribute can contain the name or identifier of a key management provider that is known to the target peer, 
+so it can be used to obtain information about encryption keys. For gateways and external schemas this can
```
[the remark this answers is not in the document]
KSDaemon: @oberstet Of course! It identifies a schema, like our `wamp.eth` — it is a known schema for e2ee, so it defines how and where keys are stored/used/received/verified.

rfc/text/advanced/ap_50_payload_passthru_mode.md
KSDaemon: Well, agree. Let's rephrase it to `mqtt` for now and add that other schemas may be added later.

rfc/text/advanced/ap_50_payload_passthru_mode.md
KSDaemon: Well is it not possible to use the same session serializer for creating the binary representation of the payload? The only limitation here is not to use json :) as it doesn't support binary types.
Maybe value `native` doesn't looks good here. As in any case payload must be serialized, but if this option is omitted then session serializer will be used (if it supports binaries)

KSDaemon: Oh... Good point: 
> this is a new meaning for "native": use the same serializer as in the enclosing WAMP session - which doesn't make sense, as there are at least 2 such sessions

So yeah, serializer needs to be specified explicitly, even if client wants to use the same one as session

rfc/text/advanced/ap_50_payload_passthru_mode.md
KSDaemon: > "wamp.eth" only allows "cbor" and "flatbuffers" - not "native"
Why not allow any serializer with binary support? E.g. why msgpack is bad?

KSDaemon:
@ecorm Yeah, I agree with @oberstet Being a member of GitHub Organization `wamp-proto` — means just that you are an active and respected member of our WAMP community, contributor, and editor.

rfc/text/advanced/ap_50_payload_passthru_mode.md
KSDaemon: So i think we should mark ppt_serializer as required for `wamp.eth` scheme, as it allows a few possible values

KSDaemon:
I've pushed some updates

KSDaemon:
So, if we already started to discuss ppt then I have another question that needs clarification:

What if a client wants to use ppt mode and provides argsList and argsKW payload at the same time?....
But the underlying wamp messages with ppt have only one [binary] arg.....

The same question is for e2ee. It should be transparent for the client code: if they previously used both argsList and argsKw and just want to add e2ee over it — nothing should be changed for them...

My first thought is to prepare a special object like: { argsList: xxxx, argsKw: yyyy } where to put client provided related payload and then serialize it if required to one [binary] and then on other peer deserialize and return as original argsList and argsKw. So for clients, that will be transparent...

I mean that it can be done in different ways. And if we do not describe it in the spec, then different implementations can be out of sync with each other %))
One decides to pack payload as an object { argsList: xxxx, argsKw: yyyy }
Other decides to pack payload as an array  [argsList, argsKw]
Another one may just disallow sending both argsList and argsKw. 

What do you think of it?

ecorm:
I have nothing to suggest from a technical perspective. Please let me know when all the technical stuff has been settled, and I'll do an English proofreading pass.

KSDaemon:
@ecorm Okay! Thnx! I'll ping you!

KSDaemon:
So I've checked Autobahn|python. Seems that e2ee implementation there already prepares payload as a hashmap. See here: https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/cryptobox.py#L221-L225

So I'll describe it in specification

rfc/text/advanced/ap_50_payload_passthru_mode.md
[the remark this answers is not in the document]
oberstet: ok, yes true, I guess I mean sth else: "so it defines how and where keys are stored/used/received/verified" => yes, it does, but to actually get at whatever key the scheme defines, _another_ field is necessary (key_id)

rfc/text/advanced/ap_50_payload_passthru_mode.md
oberstet: cool, I think this makes sense to add an AP section that describes "mqtt" (well, currently, there isn't even a proposal, only CB impl., but it is really quite straight forward - making use of the building blocks we already define in this PR)

rfc/text/advanced/ap_50_payload_passthru_mode.md
oberstet: > So yeah, serializer needs to be specified explicitly, even if client wants to use the same one as session

yes, exactly, and the reason is: the serializer is applied _before_ e2ee, and hence it cannot be translated by the router - and hence the originator would need to request one, and the responder obeying this (and support it in the first place) - however, with PubSub, the responder actually is the publisher, and the publisher doesn't want to serialize multiple times for each subscriber serializer.

IOW: I think it makes more sense to let each peer specify the serializer it did use when sending a message ...

rfc/text/advanced/ap_50_payload_passthru_mode.md
oberstet: > Why not allow any serializer with binary support? E.g. why msgpack is bad?

cbor and flatbuffers both support binaries. msgpack isn't actually bad .. we could add it. however, we need to be careful, since with e2ee each peer determines the serializer used when sending - and hence the other peer must be prepared to support all serializers. adding msgpack would require support in each client. this is different from regular WAMP - exactly because with regular WAMP the router transparently transcodes different serializers between 2 sessions involved in a WAMP action

oberstet:
> Seems that e2ee implementation there already prepares payload as a hashmap.

yes, and it also includes URI, because that means, the receiver can be sure about authenticity and integrity of URI of the WAMP action, because it is signed/encrypted. The "outside" WAMP message elements _also_ include the URI, sure, the router needs it, however, those parts are not signed or encrypted.

actually, this points to an important question which we should make a conscious decision about: the outer WAMP message element must be readable by the router, and thus cannot be e2e encrypted, however they _could_ be signed by the sending peer, which would allow the router to base its routing decisions on data which can be trusted wrt the sending peer (since it is signed by that peer)

KSDaemon:
So......
1. @oberstet @ecorm Please check the latest updates. Seems that I resolved all remarks. So I hope — we can merge it!
2. About signing:

> actually, this points to an important question which we should make a conscious decision about: the outer WAMP message element must be readable by the router, and thus cannot be e2e encrypted, however they could be signed by the sending peer, which would allow the router to base its routing decisions on data which can be trusted wrt the sending peer (since it is signed by that peer)

I think it is a great idea! We can definitely describe it in E2EE Section. I don't know what's the best: should we put it as integral part of E2EE or mark it as an additional feature on top of E2EE? From one side it is additional security feature, so we can mark it as additional feature, but from the other side — developers are lazy and not all good at security, so i suppose most will just skip it... I don't know what is the best choice. But i think going iterative is a right way. So i'll move this aspect to the future, after we describe the E2EE in spec, will have more than one implementation.

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -148,21 +151,24 @@ without guessing. The format of the value may depend on the `ppt_scheme` attribu
...
-| Attribute      | Required? | Value                       |
-|----------------|-----------|-----------------------------|
-| ppt_scheme     | Y         | mqtt                        |
-| ppt_serializer | N         | json, msgpack, cbor, native |
-| ppt_cipher     | N         | -                           |
-| ppt_keyid      | N         | -                           |
+| Attribute      | Required? | Value  |
+|----------------|-----------|--------|
+| ppt_scheme     | Y         | mqtt   |
+| ppt_serializer | N         | native |
```
oberstet, line 157: mqtt in CB support all of json, msgpack, cbor and native. pls remove the change to ppt_serializer

rfc/text/advanced/ap_50_payload_passthru_mode.md
```diff
@@ -148,21 +151,24 @@ without guessing. The format of the value may depend on the `ppt_scheme` attribu
...
+For `End-to-End Encryption` flow both peers must support chosen `ppt_serializer` regardless of their own
+session serializer.
+
 {align="left"}
 
 | Attribute      | Required? | Value                       |
 |----------------|-----------|-----------------------------|
 | ppt_scheme     | Y         | wamp.eth                    |
-| ppt_serializer | N         | cbor, flatbuffers           |
+| ppt_serializer | Y         | cbor, msgpack, flatbuffers  |
```
oberstet, line 171: AB/CB does not support msgpack. pls remove the change to ppt_serializer

oberstet (changes requested):

oberstet (approved):

oberstet:
@KSDaemon I just realized, after merging, that @ecorm offered to proof read the english, which of course would still be very welcome! we would need those changes to `rfc/text/advanced/ap_50_payload_passthru_mode.md` then in a new PR .. sorry, too quick with the merge of this PR

PR #426, opened 2022-11-14
Fixed progressive call option naming in spec

KSDaemon:
Fixed copy paste :) 
As options is called `progress` rather then `progressive_calls` (which is a feature naming)


oberstet (approved):

oberstet:
oh, right, those were bugs;)

PR #432, opened 2023-01-03
Changed event history feature status alpha->beta

KSDaemon:
I think Event history feature is well defined now in the spec and there is at least one and half implementations :)
So lets set status to `beta`. When XB will be aligned with the spec we can set it to stable, as there will be 2 broker implementations already.

### DIFF
--- a/rfc/text/advanced/ap_02_features.md
+++ b/rfc/text/advanced/ap_02_features.md
@@ -33,7 +33,7 @@ Support for advanced features must be announced by the peers which implement the
 | [Subscription Meta API](#pubsub-sub-metapi)               | beta   |   | X |   |    |   |    |
 | [Pattern-based Subscription](#pattern-based-subscription) | stable |   | X | X |    |   |    |
 | [Sharded Subscription](#pubsub-sharded-subscription)      | alpha  |   | X | X |    |   |    |
-| [Event History](#pubsub-event-history)                    | alpha  |   | X | X |    |   |    |
+| [Event History](#pubsub-event-history)                    | beta   |   | X |   |    |   |    |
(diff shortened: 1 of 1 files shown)


oberstet:
IMO "beta" should not only claims 2 implementations, but show intereoperability. eg show https://github.com/crossbario/crossbar-examples/tree/master/event-history works with the same client against 2 brokers ..

> When XB will be aligned with the spec

how is crossbar different from the spec? who is going to change it? crossbar supports event history already as alpha ...

KSDaemon:
Hm... IIRC, and also from the spec:

* **alpha** — The feature use case is still fuzzy and/or the feature definition is unclear, but there is at least a protocol-level proposal.
* **beta** — The feature use case is clearly defined and the feature definition in the spec is sufficient to write a prototype implementation. The feature definition and details may still be incomplete and change.
* **stable** — The feature definition in the spec is complete and stable and the feature use case is field proven in real applications. There are multiple, interoperable implementations.

So having interoperable implementations is required for a stable state. 

Right now, XB is a bit out of sync with the spec: it uses a second positional attribute for the `limit` option, but we moved all filtering options to kwargs. So it needs a bit of adoption. MB I'll try to find some time to update it too.

I've updated Event History Implementation in Nexus to be aligned with the spec.

So I think we're already at the `beta` stage and most likely close enough to `stable`. But I don't want to get ahead of ourselves.

oberstet (approved):

oberstet:
Ok, actually you are right: the status level descriptions are clear! And now that there is actually quite good spec text, it should be level "beta". I'll change my review and merge.

PR #434, opened 2023-01-06
Fix misspelled parameter name.

SunFulong:
Event message's parameter name should be Argument[s]Kw.


KSDaemon:
Hey @SunFulong Thnx for catching it!

SunFulong:
> Hey @SunFulong Thnx for catching it!

I'm implementing my own WAMP router and client library following the spec, it's causing my code to crash.

PR #435, opened 2023-01-06
SUBSCRIBE and UNSUBSCRIBE's request id sequential

SunFulong:
Other messages' which generate request ID are all using session scope sequential ID, should SUBSCRIBE/UNSUBSCRIBE message using sequential ID too?

### DIFF
--- a/rfc/text/basic/bp_05_publish_subscribe.md
+++ b/rfc/text/basic/bp_05_publish_subscribe.md
@@ -55,7 +55,7 @@ A Subscriber communicates its interest in a topic to a Broker by sending a `SUBS
 
 where
 
- * `Request` MUST be a random, ephemeral ID chosen by the Subscriber and used to correlate the Broker's response with the request.
+ * `Request` MUST be a sequential ID in the _session scope_, incremented by the Subscriber and used to correlate the Broker's response with the request.
  * `Options` MUST be a dictionary that allows to provide additional subscription request details in a extensible way. This is described further below.
  * `Topic` is the topic the Subscriber wants to subscribe to and MUST be a URI.
 
@@ -114,7 +114,7 @@ When a Subscriber is no longer interested in receiving events for a subscription
 
 where
 
- * `Request` MUST be a random, ephemeral ID chosen by the Subscriber and used to correlate the Broker's response with the request.
+ * `Request` MUST be a sequential ID in the _session scope_, incremented by the Subscriber and used to correlate the Broker's response with the request.
  * `SUBSCRIBED.Subscription` MUST be the ID for the subscription to unsubscribe from, originally handed out by the Broker to the Subscriber.
 
 *Example*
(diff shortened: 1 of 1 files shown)


SunFulong:
Additionally, Chapter 2.1.2 lists request ID for these messages using Session Scope IDs.

oberstet:
the spec is correct

SunFulong:
> the spec is correct

Could explain why these messages' request ID need random? and if they are random, they should not list in session scope IDs at chapter 2.1.2.

ecorm:
@SunFulong
I think you are confusing the SUBSCRIBE _request ID_ (which is session scope), with the _subscription ID_ (which is router scope).

SunFulong:
> @SunFulong
> 
> I think you are confusing the SUBSCRIBE _request ID_ (which is session scope), with the _subscription ID_ (which is router scope).

@ecorm 

No, [here](https://wamp-proto.org/wamp_latest_ietf.html#section-2.1.2-11) list SUBSCRIBE.Request is Session Scope ID, but [here](https://wamp-proto.org/wamp_latest_ietf.html#section-5.1.1-3) said its ID must be a random. I think either make the ID sequential, or remove SUBSCRIBE.Request from 2.1.2.

ecorm:
> No, [here](https://wamp-proto.org/wamp_latest_ietf.html#section-2.1.2-11) list SUBSCRIBE.Request is Session Scope ID, but [here](https://wamp-proto.org/wamp_latest_ietf.html#section-5.1.1-3) said its ID must be a random. I think either make the ID sequential, or remove SUBSCRIBE.Request from 2.1.2.

Oh I see now. Yes, it appears the following bullet point in 5.1.1 is in error and contradicts section 2.1.2.

> - Request MUST be a random, ephemeral ID chosen by the Subscriber and used to correlate the Broker's response with the request

The same error appears in [5.1.4](https://wamp-proto.org/wamp_latest_ietf.html#section-5.1.4) for UNSUBSCRIBE.

I'm 99% sure that the request ID in SUBSCRIBE and UNSUBSCRIBE messages is intended to be sequential and Session Scope, and that those were merely copy-paste errors when those sections of the standard were written.

Good catch! :+1:

ecorm:
@oberstet 
> the spec is correct

Sorry, but no, there is a contradiction between sections 2.1.2 / 5.1.1, as well as between 2.1.2 / 5.1.4 as SunFulong has pointed out.

KSDaemon:
I checked wampy, autobahn, nexus. And of course, they all take a sequential next id for these calls.
So let's reopen and merge the fix!

SunFulong:
Thank you, @KSDaemon .

I pushed a bit more modifying to make subscription chapter text like other chapters below.

oberstet:
@SunFulong sorry, seems I've been too quick / didn't read properly;)

SunFulong:
> @SunFulong sorry, seems I've been too quick / didn't read properly;)

English is not my native language, maybe my description is not clear enough, I will try to make the description clearer and more accurate in the future.

KSDaemon:
@SunFulong Don't worry. Most of us are non-native English speakers.

Issue #436, opened 2023-01-06
How to handle non-sequential request ID messages in a session?


SunFulong:
Hello,

I'm implementing my WAMP router and client library, the spec defines "IDs in the session scope MUST be incremented by 1 beginning with 1 (for each direction)", but not describe what should peer do if received non-sequential request ID.

So, what should I do if received an unexpected request ID from peer?

Thanks!

KSDaemon:
Hi! 

Well, AFAIK, nothing bad should happen. In some implementations, I'm aware of, that is not checked, so there is no problem.

We might need to clarify this in spec.

SunFulong:
How about defining a strict mode that will ABORT the session immediately if the request ID is not sequential?

ecorm:
Perhaps it should be "IDs in the session scope MUST be monotonic and SHOULD be incremented by 1 beginning with 1".

There may be use cases where a client implementation picks the request ID for an outgoing CANCEL message, but client decides to discard that message from its queue because the RPC completes before the CANCEL message is sent.

oberstet:
> but not describe what should peer do if received non-sequential request ID.

personally, and in general, I'd opt for "protocol error"

https://wamp-proto.org/wamp_latest_ietf.html#name-protocol-errors

that is harsh, but why continue talking to a broken peer? that's a risk.

in this specific case, I would expect most routers not really check that currently. should we add a UDP transport, this ID sequencing (strictly monotonic without gaps) might become actually useful .. one reason I'm -1 on changing the spec text and loosing it to "monotonic".

ecorm:
> but client decides to discard that message from its queue because the RPC completes before the CANCEL message is sent.

I should have noted that it should be harmless for the client to leave the defunct CANCEL in its transmit queue and allow it to be sent to the dealer. The dealer should just ignore it if the RPC already completed.

oberstet:
> The dealer should just ignore it if the RPC already completed.

Alright, so it's "valid behavior" for a client to send CANCEL to a router at 1k requests/sec with made up request IDs?

ecorm:
@oberstet What I'm saying is that the dealer should be able to handle the race condition where it receives a CANCEL message (due to a timeout on the caller side) after it has already sent the RESULT response. Note that the CANCEL message itself does not have its own request ID. I had raised this with #160, but I don't think it ever got resolved.

[...]

PR #437, opened 2023-01-08
Add non-sequential request ID to Protocol errors

SunFulong:
As discussed in #436 , adding message with non-sequential request ID to [protocol errors](https://wamp-proto.org/wamp_latest_ietf.html#name-protocol-errors) scenarios.

### DIFF
--- a/rfc/text/basic/bp_02_building_blocks.md
+++ b/rfc/text/basic/bp_02_building_blocks.md
@@ -262,7 +262,7 @@ The diagram below illustrates the full transport connection and session lifecycl
 
 ### Protocol errors {#protocol_errors}
 
-WAMP implementations MUST close sessions (disposing all of their resources such as subscriptions and registrations) on protocol errors caused by offending peers.
+WAMP implementations MUST abort sessions (disposing all of their resources such as subscriptions and registrations) on protocol errors caused by offending peers.
 
 Following scenarios have to be considered protocol errors:
 
@@ -287,5 +287,5 @@ Following scenarios have to be considered protocol errors:
 In all such cases WAMP implementations:
 
 1. MUST send an `ABORT` message to the offending peer, having reason `wamp.error.protocol_violation` and optional attributes in ABORT.Details such as a human readable error message.
-2. MUST close the WAMP session by disposing any allocated subscriptions/registrations for that particular client and without waiting for or processing any messages subsequently received from the peer,
+2. MUST abort the WAMP session by disposing any allocated subscriptions/registrations for that particular client and without waiting for or processing any messages subsequently received from the peer,
 3. SHOULD also drop the WAMP connection at transport level (recommended to prevent denial of service attacks)
--- a/rfc/text/basic/bp_02_building_blocks.md
+++ b/rfc/text/basic/bp_02_building_blocks.md
@@ -124,23 +124,23 @@ The following is a complete list of usage of IDs in the three categories for all
 
 **Session Scope IDs** {#session_scope_ids}
 
-* `ERROR.Request`
-* `PUBLISH.Request`
-* `PUBLISHED.Request`
 * `SUBSCRIBE.Request`
-* `SUBSCRIBED.Request`
+* `SUBSCRIBED.Request` (mirrored `SUBSCRIBE.Request`)
 * `UNSUBSCRIBE.Request`
-* `UNSUBSCRIBED.Request`
-* `CALL.Request`
-* `CANCEL.Request`
-* `RESULT.Request`
+* `UNSUBSCRIBED.Request` (mirrored `UNSUBSCRIBE.Request`)
+* `PUBLISH.Request`
+* `PUBLISHED.Request` (mirrored `PUBLISH.Request`)
 * `REGISTER.Request`
-* `REGISTERED.Request`
+* `REGISTERED.Request` (mirrored `REGISTER.Request`)
 * `UNREGISTER.Request`
-* `UNREGISTERED.Request`
+* `UNREGISTERED.Request` (mirrored `UNREGISTER.Request`)
+* `CALL.Request`
+* `RESULT.Request` (mirrored `CALL.Request`)
+* `CANCEL.Request` (mirrored `CALL.Request`)
 * `INVOCATION.Request`
-* `INTERRUPT.Request`
-* `YIELD.Request`
+* `YIELD.Request` (mirrored `INVOCATION.Request`)
+* `INTERRUPT.Request` (mirrored `INVOCATION.Request`)
+* `ERROR.Request` (mirrored original request ID)
 
 
 ## Serializers
--- a/rfc/text/basic/bp_02_building_blocks.md
+++ b/rfc/text/basic/bp_02_building_blocks.md
@@ -279,6 +279,7 @@ Following scenarios have to be considered protocol errors:
  - Receiving `REGISTERED` message, before session was established.
  - Receiving `UNREGISTERED` message, before session was established.
  - Receiving `INVOCATION` message, before session was established.
+ - Receiving message with non-sequential request ID, such as `SUBSCRIBE`, `UNSUBSCRIBE`, `PUBLISH`, `REGISTER`, `UNREGISTER`, `CALL` and `YIELD`.
  - Receiving protocol incompatible message, such as empty array, invalid WAMP message type id, etc.
  - Catching error during message encoding/decoding.
  - Any other exceptional scenario explicitly defined in any relevant section of this specification below (such as receiving a second `HELLO` within the lifetime of a session).
--- a/rfc/text/basic/bp_02_building_blocks.md
+++ b/rfc/text/basic/bp_02_building_blocks.md
@@ -97,7 +97,7 @@ These are identified in WAMP using IDs that are integers between (inclusive) **1
 
 * IDs in the *global scope* MUST be drawn *randomly* from a *uniform distribution* over the complete range [1, 2^53]
 * IDs in the *router scope* CAN be chosen freely by the specific router implementation
-* IDs in the *session scope* MUST be incremented by 1 beginning with 1 (for each direction - *Client-to-Router* and *Router-to-Client*)
+* IDs in the *session scope* MUST be incremented by 1 beginning with 1 (for each direction - *Client-to-Router* and *Router-to-Client*) {#session_scope_id}
 
 > The reason to choose the specific lower bound as 1 rather than 0 is that 0 is the null-like (falsy) value for many programming languages.
 > The reason to choose the specific upper bound is that 2^53 is the largest integer such that this integer and *all* (positive) smaller integers can be represented exactly in IEEE-754 doubles. Some languages (e.g. JavaScript) use doubles as their sole number type. Most languages do have signed and unsigned 64-bit integer types that both can hold any value from the specified range.
@@ -122,7 +122,7 @@ The following is a complete list of usage of IDs in the three categories for all
 * `INVOCATION.Registration`
 
 
-**Session Scope IDs**
+**Session Scope IDs** {#session_scope_ids}
 
 * `ERROR.Request`
 * `PUBLISH.Request`
@@ -279,7 +279,7 @@ Following scenarios have to be considered protocol errors:
  - Receiving `REGISTERED` message, before session was established.
  - Receiving `UNREGISTERED` message, before session was established.
  - Receiving `INVOCATION` message, before session was established.
- - Receiving message with non-sequential request ID, such as `SUBSCRIBE`, `UNSUBSCRIBE`, `PUBLISH`, `REGISTER`, `UNREGISTER`, `CALL` and `YIELD`.
+ - Receiving message with non-[sequential](#session_scope_id) [session scope](#session_scope_ids) request ID, such as `SUBSCRIBE`, `UNSUBSCRIBE`, `PUBLISH`, `REGISTER`, `UNREGISTER`, `CALL` and `YIELD`.
  - Receiving protocol incompatible message, such as empty array, invalid WAMP message type id, etc.
  - Catching error during message encoding/decoding.
  - Any other exceptional scenario explicitly defined in any relevant section of this specification below (such as receiving a second `HELLO` within the lifetime of a session).


oberstet (approved):

oberstet:
The added text clarifies behavior, I'm +1 on merging this. Even more explicit: the new text line could reference the section "Session Scope IDs" in this same doc. But I'm fine with the change as is as well. @ecorm @KSDaemon what do you think in light of our recent discussions?

SunFulong:
> Even more explicit: the new text line could reference the section "Session Scope IDs" in this same doc.

Sorry, I don't quite get what you mean. Should I add a reference to [that section](https://wamp-proto.org/wamp_latest_ietf.html#section-2.1.2-11) or to [sequential definition](https://wamp-proto.org/wamp_latest_ietf.html#section-2.1.2-4.3)?

BTW, I don't think messages from peer responses should be listed in the "Session Scope ID" section, their request IDs are copied from the original request, not newly generated.

oberstet:
> Should I add a reference to [that section](https://wamp-proto.org/wamp_latest_ietf.html#section-2.1.2-11) or to [sequential definition](https://wamp-proto.org/wamp_latest_ietf.html#section-2.1.2-4.3)?

yeah, that's what I meant (sorry for confusion)

> their request IDs are copied from the original request, not newly generated.

yes, that's correct: we could add a note in the section that clarifies it - I wouldn't remove the messages from the list though, they do contain session scoped IDs, just that those IDs were generated by the other peer in the first place ..

rfc/text/basic/bp_02_building_blocks.md
```diff
@@ -279,6 +279,7 @@ Following scenarios have to be considered protocol errors:
  - Receiving `REGISTERED` message, before session was established.
  - Receiving `UNREGISTERED` message, before session was established.
  - Receiving `INVOCATION` message, before session was established.
+ - Receiving message with non-sequential request ID, such as `SUBSCRIBE`, `UNSUBSCRIBE`, `PUBLISH`, `REGISTER`, `UNREGISTER`, `CALL` and `YIELD`.
```
KSDaemon, line 282: Well, here are listed only messages from client peers to the router. But I think this also applies to the reverse direction.

KSDaemon: Hm... Thinking... Are there any messages that can be initiated by router?... Seems that no. Only responses. So, sorry for the confusion.

KSDaemon (approved):

KSDaemon:
Let's add a comment about copying the `request id`. I think that would be useful. And then merge!

ecorm:
Looks fine to me. Under the subsection title, shouldn't it be:

> WAMP implementations MUST ~close~ abort sessions

"close" to me implies cleanly shutting down the session with a GOODBYE exchange, whereas "abort" implies sending an `ABORT` message with the protocol error URI and then disconnecting.

ecorm:
> Let's add a comment about copying the request id

Did you mean duplicate request IDs?

KSDaemon:
> Did you mean duplicate request IDs?

I mean the copying `Request|id` from `SUBSCRIBE` → `SUBSCRIBED` and so on

oberstet:
"duplicate" or "copy" request IDs both might suggest there is yet another, different ID involved .. which of course it isn't. a 3rd option could be: "request ID mirrored in reply". well, that's a mouthful, not 1 verb of course;)

KSDaemon:
Ou yeah! I was also thinking of `mirroring` :)

SunFulong:
Sorry for the rubbish log, I don't know only one reviewer can be requested to re-review before.

oberstet:
Don't worry about those "request review GH itches" .. we will gather feedback and reviews, we want this to be merged - once we agree on the exact details;) IOW: I'm kindly asking you for some patience, your input is much appreciated and we want to take the chance to really "fix it".

I've looked into the whole spec sections resulting again, and I just now (sorry!) one more problem:

1. @ecorm correctly noted that the wording should be "abort" rather than "close" when encountering protocol errors (this is described in the paragraph "In all such cases WAMP implementations:") - the PR reflects that now
2. BUT: the [description of ABORT](https://wamp-proto.org/wamp_bp_latest_ietf.html#name-abort-2) only has a message sequence chart that shows ABORT as being used as part of the opening of a WAMP session, and the whole ABORT message is only introduced as a subsection of "4.1 Session Establishment"

> also note that WELCOME correctly is described with its use as "completes the opening of a WAMP session", while ABORT (in the same "Session Establishment" section) is described with its use as "may abort a WAMP session". IOW: only the opening, or opening and later on as well?

If ABORT is a general alternative to GOODBYE as described in "4.2 Session Closing", and not only as part of opening a WAMP session, this should be made clear, at least by having a respective message sequence diagram for that as well.

-----

another problem is the missing explanation of why.

I mean (this is my view, lets discuss!): 

* yes, we need an ABORT message different from WELCOME in the opening of a session, as we want to be able to orderly deny a client that didn't do anything bad at the protocol level
* yes, we want a peer to "hard drop" a session by default whenever the peer did violate the protocol at a basic level, because a) continue talking to a bad (=buggy) client opens additional security risks, and b) because of [Maintaining Robust Protocols](https://datatracker.ietf.org/doc/html/draft-iab-protocol-maintenance): the Internet new-school

and we want to do the latter even though there is:

"be conservative in what you do, be liberal in what you accept from others": [Postel's law](https://en.wikipedia.org/wiki/Robustness_principle) - the Internet old-school

-----

"hard drop": https://wamp-proto.org/wamp_bp_latest_ietf.html#name-protocol-errors describes this in

> SHOULD also drop the WAMP connection at transport level (recommended to prevent denial of service attacks)

what isn't mentioned is: _if_ and implementation chooses to follow the SHOULD recommendation and "hard drop" a connection, then no ABORT message is sent at all!

SunFulong:
> I'm kindly asking you for some patience

I have chosen to use the WAMP protocol as the core of my software architecture for the next few years, so I am very concerned about this specification.

But I have a lot of work on weekdays, so the progress may be a bit slow. I usually free on weekends, please wait for me a few more days, I will definitely be back.

KSDaemon (approved):
For me this already looks good and pretty clear 👍🏻

oberstet:
@KSDaemon from my PoV, there are only 2 details lacking precision in the spec (this of course isn't a mistake in this PR, but sth we could fix within this PR as well since we are at it):

**1)**
> If ABORT is a general alternative to GOODBYE as described in "4.2 Session Closing", and not only as part of opening a WAMP session, this should be made clear, at least by having a respective message sequence diagram for that as well.

**2)**
> what isn't mentioned is: if an implementation chooses to follow the SHOULD recommendation and "hard drop" a connection, then no ABORT message is sent at all! Or is an implementation still required to follow "MUST send an ABORT message to the offending peer, having reason wamp.error.protocol_violation " ?

What's your take / interpretation?

KSDaemon:
Oh, sorry, missed it.
1) Agree! We should update the spec with appreciative diagram changes
2) Well... What is the current behavior if a peer receives an incorrect message type (like getting a second WELCOME or EVENT before the session was established)? It `MUST` send an ABORT message with `wamp.error.protocol_violation` reason and close connection. Right? Receiving non-incremental Request ID should work the same. I think it is just another type of `protocol_violation`. So I'm +1 on `MUST send an ABORT message`.

ecorm:
One thing that isn't clear to me: when using an `ABORT` during the `HELLO` handshake, should the underlying transport be disconnected afterwards, or left connected? The same question arises when using `ABORT` for protocol errors.

 > Or is an implementation still required to follow "MUST send an ABORT message to the offending peer, having reason wamp.error.protocol_violation " ?

I think we must ask ourselves what are the pros and cons of sending an ABORT in the case of protocol errors, versus terminating the transport connection.

Pros for sending abort
- Helps client developers in troubleshooting the reason why a client connection was dropped. However the same information could also be stored in router-side logs.
- Disambiguates protocol error situations versus the transport being disconnected due to network failures.

Cons for sending abort
- If transport disconnection is mandated after sending an ABORT, it can be tricky to implement in a manner that doesn't cut off the ABORT message before it is fully received by the client.
- Additional workload for the router in a DoS attack.

KSDaemon:
Good points @ecorm! 
And I still think it will be useful to send an `ABORT` message to the router. I suppose that in most high-level languages/frameworks if you issue a kind of `send` function call your app will get the control back after the message will be delivered somewhere into the OS/cache/network/buffers. So even if you close the network connection from your high-level app code, the message will already be in network subsystem queues or even on the hardware level which can handle retransmission or some other possible problems. Anyway, I don't think we should really go that deep. 
In the worst case, the abort message will just not be sent at all so the router will just close and clear the connection on its side.

oberstet:
> I think we must ask ourselves what are the pros and cons of sending an ABORT in the case of protocol errors, versus terminating the transport connection.

yes, exactly!

> Pros for sending abort

yep, that's a good list, and fwiw, I'm +1 on this option (and that would make AGREE into a message _only_ used in the opening handshake, and hence no additional sequence message chart would be needed as well)

>  if you issue a kind of send function call your app will get the control back after the message will be delivered somewhere into the OS/cache/network/buffers. 

I would argue: development is commonly done using loopback or LAN. in such a situation, when the router drops the TCP, the app will notice immediately (not via a failed WAMP call or what, but via a failed/dropped TCP transport).

If the app development is using a router over WAN, then the app developer _should_ setup client-initiated heartbeating anyways, and if so, a dropped connection will be detected in the app (on the next heartbeat send)

Only if the app development is over WAN and not using client-initiated heartbeating, the router sending ABORT could speed up the app detecting what happened.

ecorm:
It seems issue #255 isn't really resolved if we're considering the use of ABORT messages in the event of protocol errors. In that issue, @oberstet had advocated abrupt disconnection.

> I suppose that in most high-level languages/frameworks if you issue a kind of send function call your app will get the control back after the message will be delivered somewhere into the OS/cache/network/buffers.

The send function for POSIX sockets is not guaranteed to transmit all data. In C++ `boost::asio::async_write` adds the send operation to a work queue and returns immediately.

To support disconnection after ABORT, I needed to add special logic in my implementation that performs the disconnection in the `async_write` completion handler. Even then, I'm not sure of the operating system's behavior: can a socket disconnect operation cut off a network send still in progress? Should I add a delay between the end of `async_send` and socket disconnect?

KSDaemon:
Okay I understand that. But how it is done now in your implementation for cases of incorrect message types (like getting a second WELCOME)?

ecorm:
> Okay I understand that. But how it is done now in your implementation for cases of incorrect message types (like getting a second WELCOME)?

It currently sends an ABORT message and disconnects the transport socket when the networking library tells it the message has been completely sent (whatever that means at the socket/OS/network interface level). I haven't yet tested if the client fully receives the ABORT message before disconnection. Anyways. my C++ router implementation is still a work in progress, so I don't mind changing that protocol error handling behavior to match the WAMP standard.

I'm on the fence about whether ABORT messages should be sent as a result of protocol errors. I see both the advantages and disadvantages and can't decide for myself what's best. :thinking:  Should this be up to the router implementation to decide, or mandated by the standard?

ecorm:
> Okay I understand that. But how it is done now in your implementation for cases of incorrect message types (like getting a second WELCOME)?

Oh, sorry, I see now that you're asking what I currently do in my _client_ implementation. I current disconnect abruptly without sending an ABORT message.

I'm overhauling the CppWAMP client implementation at the same time that I'm writing the new router implementation. I don't mind changing the client behavior if the standard mandates that ABORT messages be sent in response to protocol errors.

The user base for CppWAMP is very small, so I don't mind making changes to better follow the standard. I'm sure that most of the clones I see in the GitHub stats are just robots. :grin:

ecorm:
> So even if you close the network connection from your high-level app code, the message will already be in network subsystem queues or even on the hardware level which can handle retransmission or some other possible problems.

I found the relevant SO question [here](https://stackoverflow.com/questions/8874021), and the recommendation in the linked [blog post](https://blog.netherlabs.nl/articles/2009/01/18/the-ultimate-so_linger-page-or-why-is-my-tcp-not-reliable) is to use the `shutdown()` system call after the message is completely sent. This sends a TCP FIN packet which causes the remote peer to close the connection after it has received the final message. The local peer then just waits for the socket to be closed by the remote host, or closes it manually after a time out.

The networking library I use [exposes `shutdown()` in its API](https://www.boost.org/doc/libs/release/doc/html/boost_asio/reference/basic_stream_socket/shutdown.html), so I can use the above procedure.

I already have to deal with this problem for `ABORT` messages sent during the `HELLO` establishment phase, so it's not any more trouble to do the same for protocol errors.

It's still not clear to me if the transport is supposed to be disconnected after an `ABORT` is sent during session establishment due to authentication failure. Should a router allow a client to attempt another WAMP session establishment via a new `HELLO`->`CHALLENGE`->`AUTHENTICATE`->`WELCOME` sequence using the same transport session?

SunFulong:
Hi all, I'm back!

I had planned to ask for a new PR after this PR was merged.

In next PR, I want to add new sections for chapters 4~6, which is an example of a complete story for each chapter. That will cover cases where the flow chart is not clear.

Maybe now is the time to submit the next PR?

oberstet:
> I already have to deal with this problem for ABORT messages sent during the HELLO establishment phase, so it's not any more trouble to do the same for protocol errors.

those are different cases. crossbar will

1. send ABORT in case of authentication failure during WAMP opening handshake
2. drop the connection upon protocol error without sending ABORT before

> It's still not clear to me if the transport is supposed to be disconnected after an `ABORT` is sent during session establishment due to authentication failure.

"sending ABORT" should be read as: the outgoing data has been put into the socket

whether that data will still hit the wire when the socket is immediately dropped (the connection closed) is not relevant, as the client must cope with TCP connections becoming unreactive anyways - however, _if_ the client is still getting data before the connection is gone, that data must be ABORT

> Even then, I'm not sure of the operating system's behavior: can a socket disconnect operation cut off a network send still in progress? 

I don't think Posix has that specified, and Linux .. not sure. There are 2 cases you can check with wireshark:

* RST vs
* FIN + ACK

https://stackoverflow.com/questions/4160347/close-vs-shutdown-socket

SunFulong:
> BUT: the [description of ABORT](https://wamp-proto.org/wamp_bp_latest_ietf.html#name-abort-2) only has a message sequence chart that shows ABORT as being used as part of the opening of a WAMP session, and the whole ABORT message is only introduced as a subsection of "4.1 Session Establishment"

I committed a new PR #439 , added an example cover ABORT message

PR #440, opened 2023-02-03
Add Octamp Client

cydrickn:


KSDaemon:
Hi @cydrickn Great to have another implementation on board! :)

KSDaemon:
Don't you mind renaming your client while it is not widely used as we already have another called the same `Swamp` swift-based?

oberstet:
oh right, fwiw I agree with @KSDaemon , the name duplication is not ideal, since it will confuse users

unfortunately, the project name and repo https://github.com/cydrickn/swamp-client have "swamp" in it. on the other hand, it's a very new project, so renaming now shouldn't be too bad ..

cydrickn:
sure, let me think for a new name for this, thanks

oberstet (approved):

KSDaemon:
@cydrickn Thank you! That's great!

Issue #265, opened 2016-11-20
Call Timeout Implementation


ericchapman:
@oberstet 
Hey Tobias,
I was trying to implement this in my Ruby library and had a few questions

```
A _timeout_ allows to *automatically* cancel a call after a specified time either at the _Callee_ or at the _Dealer_.
```

Since there is no sequence diagram, I read this as the client should perform the "CANCEL/INTERRUPT" flow when the timeout expires?  Is this correct?  If so, what should the "mode" of the cancel be?

Thanks
Eric

oberstet:
@ericchapman Hi Eric, yeah, agreed: a sequence diagram would really be useful here. Also, to be honest, the text is lacking in general:(

The general idea is that there are two possible initiators of a timeout of a pending call: the original Caller and the Router. A router that has sent out an `INVOCATION` message to a Callee and does not a receive a `YIELD` within a time can fail the call and return a `CALL_ERROR` to the Callee. Or the Callee that has sent a `CALL` message to the Router and does not receive either a `CALL_RESULT` or `CALL_ERROR` within a time can fail the call and raise an exception in application code.

I think the former (the Router handling the timeout logic) is definitely useful. A client can simply set a value for call timeout, and the Router handles the annoying part. Apparantly, I originally thought there were reasons for the second variant (the Caller handling the timout logic), but right now, mmh;)

ericchapman:
I agree with the router handling it.  I had it implemented as a timer going off on the client and the timer would send the cancel which would take it round trip and then eventually the client would get the error back.  Seems like overkill

oberstet:
Yep, having the router take the burden of maintaining timers also is reusable across client libraries (does not need to be implemented in each and every client library).

Another thing that needs agreement: the error URI within the `CALL_ERROR` the Dealer sends back to the Client upon a timeout (on the `INVOCATION` message sent to the Callee and no `YIELD` being received in time).

Yet another thing: do we send an `INTERRUPT` to the Callee to inform it that there is no need to produce a result anymore (since the call was already timed out). Like with explicit call cancel: https://github.com/wamp-proto/wamp-proto/blob/master/rfc/text/advanced/ap_rpc_call_canceling.md

And finally: we can't keep the state for a canceled call around for infinite time, and hence, we'll loose the invocation request ID, and therefor, should a result come back from the Callee after the timeout, we can't distinguish that from a mere bogus invocation request ID.

[...]

PR #441, opened 2023-02-06
Specify Call Timeouts More Clearly

### LINKED ISSUE #155: Message flow is not specified in Call Timeouts
It's not obvious what the sequence of messages should look like for [Call Timeouts](https://github.com/tavendo/WAMP/blob/master/spec/advanced.md#call-timeouts). My guess is that its identical to the sequence shown in [Canceling Calls](https://github.com/tavendo/WAMP/blob/master/spec/advanced.md#canceling-calls), with the exception that the caller does not send a `CANCEL` message. In any case, the intended flow should be made explicit for call timeouts.

ecorm:
Resolves #155

Pinging those who commented in the related issue: @estan @gammazero @oberstet


oberstet (approved):

ecorm:
I think this also resolves #265 . Pinging @ericchapman who raised that issue.

KSDaemon (approved):
I think it is clear! 👍🏻

KSDaemon:
Let's merge? @ecorm

oberstet:
@KSDaemon I've updated the branch to master and merged, as @ecorm is the author and we both reviewed/ack'ed already .. hope this is good for all

KSDaemon:
Absolutely! 👍🏻

ecorm:
Thank you, @oberstet  and @KSDaemon !

PR #444, opened 2023-02-16
Define all AA-related error URIs systematically

SunFulong:
Adding two error URIs, as description in PR:

1. A *Router* requires authentication, but *Client* does not provide authentication information in *HELLO* message;
2. A *Router* requires authentication, but *Client* didn't provide valid authentication information in *CHALLENGE* message.

### DIFF
--- a/rfc/text/advanced/ap_48_uris.md
+++ b/rfc/text/advanced/ap_48_uris.md
@@ -2,15 +2,10 @@
 
 WAMP pre-defines the following error URIs for the **Advanced Profile**. WAMP peers SHOULD only use the defined error messages.
 
-A *Router* requires authentication, but *Client* does not provide authentication information in *HELLO* message
+A *Router* requires authentication, but *Client* does not provide matching informations in *HELLO* message or does not provide valid informations in *CHALLENGE* message
 
 {align="left"}
-        wamp.error.authentication_required
-
-A *Router* requires authentication, but *Client* didn't provide valid authentication information in *CHALLENGE* message
-
-{align="left"}
-        wamp.error.authentication_failure
+        wamp.error.authentication_failed
 
 A *Dealer* or *Callee* canceled a call previously issued
 
--- a/rfc/text/advanced/ap_48_uris.md
+++ b/rfc/text/advanced/ap_48_uris.md
@@ -2,7 +2,17 @@
 
 WAMP pre-defines the following error URIs for the **Advanced Profile**. WAMP peers SHOULD only use the defined error messages.
 
-A *Dealer* or *Callee* canceled a call previously issued
+A *Router* requires authentication, but *Client* does not provide authentication information in *HELLO* message
+
+{align="left"}
+        wamp.error.authentication_required
+
+A *Router* requires authentication, but *Client* didn't provide valid authentication information in *CHALLENGE* message
+
+{align="left"}
+        wamp.error.authentication_failure
+
+A *Dealer* or *Callee* canceled a call previously issued
 
 {align="left"}
         wamp.error.canceled


KSDaemon:
There are two related auth error URIs in the basic profile described here: https://wamp-proto.org/wamp_latest_ietf.html#name-basic-profile-uris

* wamp.error.not_authorized
* wamp.error.authorization_failed

I think we should kind of synchronize them. Because they are mostly about the same.

SunFulong:
> I think we should kind of synchronize them. Because they are mostly about the same.

Yes, I agree too, I didn't notice the `join` in the basic profile.

ecorm:
> There are two related auth error URIs in the basic profile described here: https://wamp-proto.org/wamp_latest_ietf.html#name-basic-profile-uris
>
>    - wamp.error.not_authorized
>    - wamp.error.authorization_failed

Please don't confuse _authorization_ with _authentication_! Autobahn [uses](https://github.com/crossbario/autobahn-python/blob/01165cf27ada0c246cc23a1e5aac07a6b6839bdc/autobahn/wamp/protocol.py#L581) `wamp.error.cannot_authenticate` for authentication failures, which is not documented anywhere in the WAMP spec.

SunFulong:
> Please don't confuse authorization with authentication!

You are right, now reopen, please review again.

ecorm:
There should only be one URI for authentication failures. If the router wants to provide additional information on why the authentication failed, it can pass a string to the `ABORT.Details.message` option.

Note that it's rarely ever a good idea to provide any information to the client about why an authentication failed. This information could be leveraged by an attacker (e.g. that a certain username exists on the system).

SunFulong:
> There should only be one URI for authentication failures.

I designed two different scenarios:
1. (Non-anonymous) `router` receives a `HELLO` message that does not contain authentication information;
2. The `router` received an `AUTHENTICATE` message without valid authentication information.

Both scenarios all reply with `wamp.error.authentication_failure`?

SunFulong:
OK, combined.

ecorm:
Perhaps it would be better to use `wamp.error.cannot_authenticate` seeing that it's already in use by Crossbar/Autobahn.

ecorm:
> Perhaps it would be better to use `wamp.error.cannot_authenticate` seeing that it's already in use by Crossbar/Autobahn.

I may have been mistaken. `wamp.error.cannot_authenticate` appears to be an error URI used locally by an Autobahn client. I don't think that URI is emitted by the Crossbar router.

I found `wamp.error.authentication_failed` [here](https://github.com/crossbario/autobahn-python/blob/01165cf27ada0c246cc23a1e5aac07a6b6839bdc/autobahn/wamp/exception.py#L189) in the Autobahn|Python repo. But its description says:

> Something failed with the authentication itself, that is, authentication could not run to end.

We need an URI that means "authentication rejected". I'm going to keep digging in the Crossbar/Autobahn code to see what they use.

@oberstet , @meejah , please feel free to indicate which URI is sent by Crossbar when an authentication is rejected (e.g. invalid AUTHENTICATION signature).

ecorm:
The Github online search feature is pretty useless, so I downloaded the Crossbar and Autobahn repos to search locally with an IDE. It turns out that Crossbar uses the `wamp.error.not_authorized` URI in ABORT messages when an authentication request is denied.
See https://github.com/crossbario/crossbar/blob/427cf58e42014499a623d4c17e06c6e3440273f9/crossbar/router/auth/ticket.py#L142
and https://github.com/crossbario/autobahn-python/blob/01165cf27ada0c246cc23a1e5aac07a6b6839bdc/autobahn/wamp/types.py#L189

It appears `wamp.error.not_authorized` has a dual meaning in Crossbar:
- An operation (CALL, PUBLISH, etc) is not authorized (via the dynamic authorizer, for example)
- An authentication attempt was denied.

`wamp.error.authentication_failure` in Crossbar means that the authentication operation itself failed. For example, something went wrong with the dynamic authenticator.

There appears to be broken symmetry when it comes to authorization/authentication and error URIs:

Condition | Authorization Error URI | Authentication Error URI
-- | -- | --
Denied | wamp.error.not_authorized | wamp.error.not_authorized
Failed | wamp.error.authorization_failed  | wamp.error.authentication_failed

oberstet:
IMO we should differentiate systematically between these cases

1. authentication
2. authorization

X

A. required by router, but not provided by client
B. required by router, provided by client, but client credentials rejected by router
C. required by router, and client provided, but verification failed router-wise (eg the dynamic authenticator/authorizer code, which is user code, not router code, in general)

*so we need 6 URIs in total*

> this is likely not following a complete systematic scheme in Crossbar.io/Autobahn currently - history. we should take the chance and clean it up now, in the spec, once and for all.

KSDaemon:
> B. required by router, provided by client, but client credentials rejected by router
C. required by router, and client provided, but verification failed router-wise

I don't clearly understand what `rejected by router` mean. @oberstet Can you explain?
If the router has a kind of static credentials lookup table, then rejecting due to incorrect credentials provided is still a verification fail, no matter static or dynamic.

From one side: we can create an error URI for every possible case, but from the other: do they all will really be used?
From my experience and what I see all around: it is just a kind of `generalized` auth error and that is all. 

So I feel to create just a few high-view distinguished errors: one for authorization, one for authentication, and so on, without more granularity (auth fail because of this, because of that, etc). Just because I think no one cares much. But maybe I'm wrong.
Just my imho :)

oberstet:
> I don't clearly understand what `rejected by router` mean. @oberstet Can you explain?

e.g. the client gave a wrong password (or similar). so it was **denied**

this is different from: the dynamic authenticator procedure failed (raised an exception), or was configured but isn't online. so it was **failed**

if the client did not give any password (or similar), it is **denied** (or, with what I proposed **missing**)

KSDaemon:
I see, thnx!

Well, for me cases: 

* gave the wrong password or
* Not gave password at all  

Are the same: authentication failed.

The only case that might be different — is if the dynamic authenticator is misbehaving somehow (offline, exceptions, etc).

ecorm:
It appears we have to choose if we want to be minimalist with error URIs related to authentication/authorization, or to be prolific. I just want there to be some logic to it, and avoid instances where `authorization` in the URI applies to both authorization and authentication. I'll try to summarize the continuum of possibilities here:

Condition | A (minimalist) | B | C (prolific)
-- | -- | -- | --
Authentication Denied | wamp.error.auth_denied | wamp.error.auth_denied |  wamp.error.authentication_denied
Authentication Failed | wamp.error.auth_denied | wamp.error.auth_failed | wamp.error.authentication_failed
Authorization Denied | wamp.error.auth_denied | wamp.error.auth_denied | wamp.error.authorization_denied
Authorization Failed | wamp.error.auth_denied | wamp.error.auth_failed | wamp.error_authorization_failed

Add to that the Authentication Missing condition which can either be combined with the Authentication Denied condition, or have its own distinct URI.

Note that `wamp.error.auth_denied` could be shortened to just `wamp.error.denied`.

If column A were chosen we'd have:

- `wamp.error.denied`: The authentication was denied (e.g. bad credentials), or the user is not authorized to perform the operation, or the authentication/authorization operation itself failed.

If column B were chosen we'd have:

- `wamp.error.auth_denied`: The authentication was denied (e.g. bad credentials), or the user is not authorized to perform the operation.
- `wamp.error.auth_failed`: The authentication or authorization operation itself failed (e.g. dynamic authentication/authorizer failure).

If column C were chosen we'd have:

- (optionally) `wamp.error.authentication_missing`: Authentication information is missing.
- `wamp.error.authentication_denied`: The authentication was denied (e.g. bad credentials or missing information).
- `wamp.error.authentication_failed`: The authentication operation itself failed (e.g. dynamic authenticator failure).
- `wamp.error.authorization_denied`: The user is not authorized to perform the operation.
- `wamp.error.authorization_failed`: The authorization operation itself failed (e.g. dynamic authorizer failure).

With the minimalist approach, there are few error URIs to implement in our code, but the URIs have loose, overloaded meanings. With the prolific approach, there are more URIs to implement in our code, but the URIs have precise, singular meanings.

KSDaemon:
Hah) @ecorm In your table you already compressed several cases into one. Like: `wamp.error.authentication_denied`: it can be raised because no credentials were provided or they were provided but incorrect :)

So If to choose between these A/B/C variants — I like `C` as the most descriptive. But there are still only 4 different URIs and not the 6+

oberstet:
> I like C as the most descriptive. But there are still only 4 different URIs and not the 6+

I agree, I'm +1 on C with 4 cases. D would have been 6 cases, but ok, kick it, 4 is enough.

I'm definitely -1 on _not_ using different URIs for authentication vs authorization (both A and B)

ecorm:
> KSDaemon: In your table you already compressed several cases into one. Like: wamp.error.authentication_denied: it can be raised because no credentials were provided or they were provided but incorrect :)

> oberstet: D would have been 6 cases, but ok, kick it, 4 is enough.

Please see the note below the table that says:

> Add to that the Authentication Missing condition which can either be combined with the Authentication Denied condition, or have its own distinct URI.

Unless I'm mistaken, there is no such thing as an Authorization Missing condition, so there could be 5 error conditions maximum.

As I've already mentioned to SunFulong above, if a client doesn't not know that they're supposed to provide authentication information, then I wouldn't trust it and would not provide it any more information than necessary other than "denied".

> KSDaemon: I like C as the most descriptive

> obserstet: I agree, I'm +1 on C with 4 cases. D would have been 6 cases, but ok, kick it, 4 is enough.

That has the advantage that WAMP's `wamp.error.authorization_failed` and Crossbar's `wamp.error.authentication_failed` don't need to be deprecated.

I think `wamp.error.not_authorized` should be deprecated in favor of `wamp.error.authorization_denied`. We'd also have to add `wamp.error.authentication_denied`.

ecorm:
Here's the table again with the Authentication Missing condition included:

Condition | A (minimalist) | B | C (prolific) | D (most prolific)
-- | -- | -- | -- | --
Authentication Missing | wamp.error.auth_denied | wamp.error.auth_denied |  wamp.error.authentication_denied |  wamp.error.authentication_missing
Authentication Denied | wamp.error.auth_denied | wamp.error.auth_denied |  wamp.error.authentication_denied |  wamp.error.authentication_denied
Authentication Failed | wamp.error.auth_denied | wamp.error.auth_failed | wamp.error.authentication_failed |  wamp.error.authentication_denied
Authorization Denied | wamp.error.auth_denied | wamp.error.auth_denied | wamp.error.authorization_denied |  wamp.error.authentication_denied
Authorization Failed | wamp.error.auth_denied | wamp.error.auth_failed | wamp.error_authorization_failed |  wamp.error.authentication_denied

KSDaemon:
Emmm..... @ecorm The `D` column seems to be a bit copy-pasted :)

oberstet:
> Emmm..... @ecorm The `D` column seems to be a bit copy-pasted :)

yeah, also noticed;)

anyways: I'm +1 on C

it keeps distinction between authentication vs authorization and denied vs failed, and thus has 4 values applied in 5 cases. I like differentiate between "case/condition" and "error URI"

I'm cool with 4 error URIs, also, because I agree with

> if a client doesn't not know that they're supposed to provide authentication information, then I wouldn't trust it and would not provide it any more information than necessary other than "denied".

ecorm:
> Emmm..... The D column seems to be a bit copy-pasted :)

Oops, fixed.

KSDaemon:
Just for the record: I'm +1 on `C` or `D`.

SunFulong:
I'd like `wamp.error.authentication_required` instead of `wamp.error.authentication_missing` if router running at non-anonymous (strict auth) mode.

oberstet:
@SunFulong _if_ we go for D, then I agree, "authentication_required" is more precise, since some piece of information can be "missing" (not present) without being "required" when it is "optional". but for authentication the information isn't optional.

so _if_ we go for D (and use 3 different URIs not just 2 for authentication), in summary:

* `wamp.error.authentication_required`: router demands authentication, but client did use "anonymous" or did not provide all necessary information for authentication
* `wamp.error.authentication_failed`: the router-side authentication failed (technically), and whether the client would have been authenticated could not be determined (it might have or it might not have)
* `wamp.error.authentication_denied`: the client provided authentication information, and the router-side authentication ran properly (technically), but it determined the client isn't authenticated successfully (eg "wrong password")

the URIs for `wamp.error.authorization_*` would use the same suffixes

ecorm:
I agree that `wamp.error.authentication_required` is better than `wamp.error.authentication_missing` and I'm not against its inclusion.

> the URIs for wamp.error.authorization_* would use the same suffixes

Does that also mean `wamp.error.authorization_required` would be added? If so, how could this error occur?

oberstet:
> Does that also mean `wamp.error.authorization_required` would be added? If so, how could this error occur?

This is a very good question!

This depends on the kind of access control mechanism a router implements.

E.g. Crossbar.io currently only implements RBAC / MAC. With that mechanism, access control is centrally mandated by the node administrator, either via static node configuration of via dynamic authorizer.

Since access is bound to the triple `AC = (realm, role, uri-pattern, action) -> true|false`, there are (well, would, since it needs adjusts to new URIs we agree on) only these URIs used:

* `wamp.error.authorization_failed`: the (dynamic) authorizer failed (a static authorizer "can not" fail technically .. unless there is a bug)
* `wamp.error.authorization_denied`: the authorization is denied

However, I am thinking about adding a **Capability based Access Control** for WAMP in Crossbar.io.

With that, the originating peer of an action must provide a "Capability" for the action it wants to perfom, when it wants to perform that action.

If it doesn't provide such a Capability while the URI for the action matches for **Capability based Access Control**  in Crossbar.io, the router then would deny:

* `wamp.error.authorization_required`: an authorization by the client (a capability) is required to perform the action

oberstet:
I would also like to add (because it "fits" this discussion context):

Authorization and Access Control can be enforced (or should be enforceable) by a Router even when the application payload is end-to-end encrypted (https://github.com/wamp-proto/wamp-proto/pull/419) - IOW: it is orthogonal, even though a client authorized to call some procedure might not have the data encryption key to decrypt the result

And it should be possible to use either RBAC-MAC or CAP-MAC with E2E.

ecorm:
> However, I am thinking about adding a Capability based Access Control for WAMP in Crossbar.io.

I read-up on Cabability-based security, and what you're saying makes sense. Yes, let us standardize `wamp.error.authorization_required` now so that router implementations don't make up their own diverging error URIs.

oberstet:
I've consolidated our discussion above into a new proposal here https://github.com/wamp-proto/wamp-proto/pull/452

@ecorm wanted to request your review on this PR, but for whatever reasons, GH doesn't allow me to. strange. In any case, I've invited you for write access to this repo.

@SunFulong please also have a look if you find time. it should include/reflect your original change

ecorm:
@oberstet I thank you for the invitation, but I'm not comfortable having write access to this repo. In the future, if you don't mind, you can just @mention me in the PR. That's how KSDaemon did it when he wanted my review.

PR #445, opened 2023-02-21
Consolidate progress call results stuff and clarify progressive calls language

ecorm:
All mentions of Progressive Call Results have been consolidated to its own sub-section. This makes it easier to understand the fundamentals of Progressive Calls before mixing in Progressive Call Results.

I added the missing requirement that Callees must announce the `call_cancelling` feature.

I also took the liberty of enhancing the English grammar and style while leaving the technical requirements intact.

### DIFF
--- a/rfc/text/advanced/ap_10_rpc_progressive_calls.md
+++ b/rfc/text/advanced/ap_10_rpc_progressive_calls.md
@@ -1,19 +1,15 @@
 ## Progressive Calls {#rpc-progressive-calls}
 
-A *Caller* may issue a progressive call. This can be useful in few cases:
+A *Caller* may issue a progressive call. This can be useful in a few cases:
 
-* Payload is too big to send it whole in one request. E.g. uploading a file.
-* Payload is long term-based, and it is required to start processing it earlier.
-* Partial results should be made available to the caller as quickly as possible. This can be accomplished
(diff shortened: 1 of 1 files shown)


oberstet:
LGMT! +1

oberstet (approved):

ecorm:
I just thought of the following corner case: What happens when a progressive call requests progressive results, but in a subsequent progressive call it no longer requests progressive results?

ecorm:
One thing that's missing is what happens when a CALL message uses a request ID for a progressive call that no longer exists (due to the callee unregistering or leaving) and the caller has not yet received the `wamp.error.cancelled` error. I don't think aborting with `wamp.error.protocol_violation` would be appropriate here, because the race condition is outside the caller's control.

How about `wamp.error.call_expired`?

Note that when progressive calls are supported, a router can no longer reject CALL request IDs that are not sequential. The router must allow the caller to use an old request ID that may reference a recently expired progressive call. It would be too costly for a router to remember all previous progressive call request IDs.

KSDaemon (approved):
Great! 👍🏻

KSDaemon:
Merging this. Let's discuss the rest in other issues.

ecorm:
> Merging this. Let's discuss the rest in other issues.

@KSDaemon please see issue #448 I created to continue the discussion.

PR #449, opened 2023-02-23
Frozen Progressive Call options and Progressive Call Results timeouts

ecorm:
Pinging @oberstet and @KSDaemon who commented in the related issues.

### DIFF
--- a/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
+++ b/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
@@ -339,3 +339,8 @@ Note: Any `ERROR` returned by the *Callee*, in response to the `INTERRUPT`, is i
 A *Callee* that does not support progressive results SHOULD ignore any `INVOCATION.Details.receive_progress` flag.
 
 A *Callee* that supports progressive results, but does not support call canceling is considered by the *Dealer* to not support progressive results.
+
+
+**Timeouts**
+        
+When the *Call Timeouts* feature is used in combination with Progressive Call Results, the `CALL.Options.timeout|int` option shall represent the time limit between the initial call and the first result, and between results thereafter.
--- a/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
+++ b/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
@@ -343,4 +343,14 @@ A *Callee* that supports progressive results, but does not support call cancelin
 
 **Timeouts**
         
-When the *Call Timeouts* feature is used in combination with Progressive Call Results, the `CALL.Options.timeout|int` option shall represent the time limit between the initial call and the first result, and between results thereafter.
+When the *Call Timeouts* feature is used in combination with Progressive Call Results, the `CALL.Options.timeout|integer` option shall represent the time limit between the initial call and the first result, and between results thereafter.
+
+For Dealer-initiated timeouts, this corresponds to
+- the time between receipt of the `CALL` message and receipt of the first `YIELD` message, and,
+- the time between received `YIELD` messages thereafter.
+
+For Callee-initiated timeouts, this corresponds to
+- the time between receipt of the `INVOCATION` message and acquisition of the first result, and,
+- the time between acquisition of successive results thereafter.
+        
+Note that for progressive results, the timeout value does _not_ correspond to the duration of the complete call from initiation to the final result. The rationale for this is that it would be unfeasible to compute a reasonable timeout value for a call having a non-deterministic number of progressive results.
--- a/rfc/text/advanced/ap_10_rpc_progressive_calls.md
+++ b/rfc/text/advanced/ap_10_rpc_progressive_calls.md
@@ -259,3 +259,15 @@ to be matched pairs.
 
 Because they are part of the same call operation, the request ID is the same in all `CALL`,
 `INVOCATION`, `YIELD`, and `ERROR` messages in the above exchange.
+
+
+**Stickiness of Progressive Call Options**
+
+Except for `progress`, items in the `Options` dictionary of the initiating progressive `CALL` shall be effective for
+the entirety of the progressive call request. Except for `progress`, all options in subsequent progressive calls
+(within the same overall request) shall be ignored by the *Dealer*. Except for `progress`, items in the `Details`
+dictionary of the corresponding `INVOCATION` message shall be based on the initiating progressive `CALL` only.
+  
+For example, if `disclose_me=true` was specified in the initiating progressive call, all subsequent progressive calls
+(within the same request) shall be considered by the *Dealer* to have `disclose_me=true` as well. That is, all
+`INVOCATION` messages associated with the overall request shall contain caller identify information.
--- a/rfc/text/advanced/ap_10_rpc_progressive_calls.md
+++ b/rfc/text/advanced/ap_10_rpc_progressive_calls.md
@@ -261,9 +261,9 @@ Because they are part of the same call operation, the request ID is the same in 
 `INVOCATION`, `YIELD`, and `ERROR` messages in the above exchange.
 
 
-**Establishment of Options in Progressive Calls**
+**Freezing of Options in Progressive Calls**
 
-Except for `progress`, items in the `Options` dictionary of the initiating progressive `CALL` shall be effective for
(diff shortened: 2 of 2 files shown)


rfc/text/advanced/ap_09_rpc_progressive_call_results.md
```diff
@@ -339,3 +339,8 @@ Note: Any `ERROR` returned by the *Callee*, in response to the `INTERRUPT`, is i
 A *Callee* that does not support progressive results SHOULD ignore any `INVOCATION.Details.receive_progress` flag.
 
 A *Callee* that supports progressive results, but does not support call canceling is considered by the *Dealer* to not support progressive results.
+
+
+**Timeouts**
+        
+When the *Call Timeouts* feature is used in combination with Progressive Call Results, the `CALL.Options.timeout|int` option shall represent the time limit between the initial call and the first result, and between results thereafter.
```
oberstet, line 346: does that mean, given a call timeout T:

* `t(YIELD_1 or ERROR_1 received from callee) - t(CALL received from caller) < T AND`
* `t(YIELD_n+1 or ERROR_n+1 received from callee) - t(YIELD_n received from callee) < T`

?

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -259,3 +259,15 @@ to be matched pairs.
 
 Because they are part of the same call operation, the request ID is the same in all `CALL`,
 `INVOCATION`, `YIELD`, and `ERROR` messages in the above exchange.
+
+
+**Stickiness of Progressive Call Options**
+
+Except for `progress`, items in the `Options` dictionary of the initiating progressive `CALL` shall be effective for
```
oberstet, line 266: a positive/concrete list of items in `Options` would be better IMO. a list with the options that are indeed "effective for the call".

oberstet (changes requested):

oberstet:
regarding "Stickiness of Progressive Call Options": this is about progressive call arguments, not progressive call results, right?

if so, we need to talk about this list in `CALL` and `INVOCATION` messages.

eg going through the list of `INVOCATION.Details` in autobahn-python

https://github.com/crossbario/autobahn-python/blob/01165cf27ada0c246cc23a1e5aac07a6b6839bdc/autobahn/wamp/message.py#L5202

```
        'request', => OK
        'registration', => OK
        'args', => OK
        'kwargs', => OK
        'payload', => OK (payload passthru)
        'timeout', => ???
        'receive_progress', => FIXED from 1st
        'caller', => FIXED from 1st
        'caller_authid', => FIXED from 1st
        'caller_authrole', => FIXED from 1st
        'procedure', => FIXED from 1st
        'transaction_hash', => ???
        'enc_algo', => FIXED from 1st
        'enc_key', => FIXED from 1st
        'enc_serializer', => FIXED from 1st
        'forward_for', => ???
```

* `timeout`: callee-side invocation timeout - should that be fixed from 1st CALL?
* `transaction_hash`: crossbar specific app payload transaction hash .. I'd keep it fixed from 1st CALL
* `forward_for`: router-to-router links .. if progressive call arguments might travel different node routes, this would need to be variable. I'd keep it fixed, which would imply: all progressive arguments of a single call must travel the same nodes route

rfc/text/advanced/ap_09_rpc_progressive_call_results.md
KSDaemon: I understand it like you wrote

rfc/text/advanced/ap_10_rpc_progressive_calls.md
KSDaemon: Well, why not. There are not so many of them. So the reader can see them here in place and doesn't have to jump to other spec sections. The one downside is: if we later add new options - I'm sure everyone will forget to add them here.

rfc/text/advanced/ap_10_rpc_progressive_calls.md
```diff
@@ -259,3 +259,15 @@ to be matched pairs.
...
+the entirety of the progressive call request. Except for `progress`, all options in subsequent progressive calls
+(within the same overall request) shall be ignored by the *Dealer*. Except for `progress`, items in the `Details`
+dictionary of the corresponding `INVOCATION` message shall be based on the initiating progressive `CALL` only.
```
KSDaemon, line 269: By the way, it might be worth noting that any options (besides `progress`) can be omitted altogether in subsequent calls.

KSDaemon:
> timeout: callee-side invocation timeout - should that be fixed from 1st CALL?

I think `YES`. @ecorm described how it works for progressive calls.

rfc/text/advanced/ap_09_rpc_progressive_call_results.md
ecorm: @oberstet Assuming that the timeout is handled by the Dealer (and not delegated to the Callee), then yes, that's what I meant. There should be no need to add any timestamps to messages.

Would you like me to add your equations so that it's more formal?

rfc/text/advanced/ap_10_rpc_progressive_calls.md
ecorm: I'm on the fence for that idea. While it would indeed be convenient to have a list right there, it's going to be a bit of a maintenance problem when new CALL options are added and someone forgets to update that list. Also consider the possibility of the spec being split up into more "modular" RFCs (if that's something you're willing to consider).

We could add a warning that the list may or may not be exhaustive.

If @oberstet insists on that list, then I'll add it. It's no big deal to me either way.

rfc/text/advanced/ap_10_rpc_progressive_calls.md
ecorm: Agreed. I'll add that note.

ecorm:
> regarding "Stickiness of Progressive Call Options": this is about progressive call arguments, not progressive call results, right?

That's correct. The feature is named Progressive Calls in the spec, so I try to be consistent with the existing nomenclature. Would "Option Stickiness in Progressive Calls" be clearer?

> timeout: callee-side invocation timeout - should that be fixed from 1st CALL?

That's what I propose, yes. Except for `progress`, I think it would be preferable that there be as few exceptions as possible to the "every option in the first progressive call is sticky" rule.

Consider the case where Progressive Calls are combined with Progressive Call Results. If timeout values can change at every progressive CALL, I would make the implementation of Progressive Call Results more complex.

> -`transaction_hash`: crossbar specific app payload transaction hash .. I'd keep it fixed from 1st CALL
> -`forward_for`: router-to-router links .. if progressive call arguments might travel different node routes, this would need to be variable. I'd keep it fixed, which would imply: all progressive arguments of a single call must travel the same nodes route

I'm not familiar with those features, so I'll bow to your wisdom and will follow your recommendations.

rfc/text/advanced/ap_10_rpc_progressive_calls.md
ecorm: Not having to repeat (and not being able to change) options is more in line with the idea of a media stream where you set things up initially, and just keep sending more data thereafter.

ecorm: > Not having to repeat (and not being able to change) options is more in line with the idea of a media stream where you set things up initially, and just keep uploading more data thereafter.

I think that comment (or similar) deserves to be added to the spec as rationale for _why_ progressive call options are sticky.

KSDaemon: yeah, agree!

rfc/text/advanced/ap_09_rpc_progressive_call_results.md
oberstet: > Would you like me to add your equations so that it's more formal?

for my taste, yes, something a bit more formal or descriptive. e.g.

* this is about dealer-side/enforced callee timeouts
* the timeout applies for "time between successive progressive results" (_not_ a timeout for the complete call from start to last final result)
* a consequence of the last: the dealer needs to track 1 timeout instance per received progressive result

ecorm:
@oberstet, @KSDaemon: made all requested changes. Please review.

KSDaemon (approved):
Looks good! 👍🏻

oberstet (approved):

PR #452, opened 2023-02-27
[WIP] AA related URIs

oberstet:
this PR is supposed to consolidate and cleanup the discussion from https://github.com/wamp-proto/wamp-proto/pull/444 and supersedes the latter

The PR adds:
1. the URIs and descriptions from the discussion above - hopefully
2. it also adds `no_auth_method`, `no_such_realm`, `no_such_role` and `no_such_principal`

I have done 2. after reviewing this whole new section and after reviewing auth code in crossbar/autobahn code. It is necessary there, and I think it makes sense, *but I can remove that again and only keep 1!*. Let me know.

### DIFF
--- a/rfc/text/advanced/ap_48_uris.md
+++ b/rfc/text/advanced/ap_48_uris.md
@@ -19,7 +19,7 @@ The *Client* attempted to authenticate for a non-existing *Role* (`authrole|stri
 {align="left"}
         wamp.error.no_such_role
 
-The *Client* authenticated for a *Principal* (`authid|string`) that does not (or no longer) exists.
+The *Client* authenticated for a non-existing *Principal* (`authid|string`).
 
 {align="left"}
         wamp.error.no_such_principal
--- a/rfc/text/advanced/ap_48_uris.md
+++ b/rfc/text/advanced/ap_48_uris.md
@@ -46,7 +46,7 @@ The *Principal* is not authorized to perform such *Action*.
 {align="left"}
         wamp.error.authorization_denied
 
-Authorization of the *Principal* for the *Action* could not be determined as it failed technically at run-time, and the *Action* is therefor rejected ("fail secure operation").
+Authorization of the *Principal* to perform the given *Action* was rejected due to a technical runtime failure ("fail secure" operation).
 
 {align="left"}
         wamp.error.authorization_failed
--- a/rfc/text/advanced/ap_48_uris.md
+++ b/rfc/text/advanced/ap_48_uris.md
@@ -2,6 +2,62 @@
 
 WAMP pre-defines the following error URIs for the **Advanced Profile**. WAMP peers SHOULD only use the defined error messages.
 
+## Authentication
+
+No authentication method the *Client* offered is accepted.
+
+{align="left"}
+        wamp.error.no_auth_method
+
+The *Client* authenticated for a *Realm* (`realm|string`) that does not (or no longer) exists.
+
+{align="left"}
+        wamp.error.no_such_realm
+
+The *Client* authenticated for a *Role* (`authrole|string`) that does not (or no longer) exists.
+
+{align="left"}
+        wamp.error.no_such_role
+
+The *Client* authenticated for a *Principal* (`authid|string`) that does not (or no longer) exists.
+
+{align="left"}
+        wamp.error.no_such_principal
+
+The authentication as presented by the *Client* is denied (e.g. "wrong password").
+
+{align="left"}
+        wamp.error.authentication_denied
+
+The authentication of the *Client* failed technically, and the *Client* is therefor rejected ("fail secure").
+
+{align="left"}
+        wamp.error.authentication_failed
+
+The *Client* did not authenticate (at all) and a successful (non-anonymous) authentication is required.
+
+{align="left"}
+        wamp.error.authentication_required
+
+## Authorization
+
+The *Client* is not authorized to perform such action.
+
+{align="left"}
+        wamp.error.authorization_denied
+
+Authorization of the *Client* for the action could not be determined as it failed technically.
+
+{align="left"}
+        wamp.error.authorization_failed
+
+Authorization by the *Client* ("capability") is required for each single action.
+
+{align="left"}
+        wamp.error.authorization_required
+
+## Remote Procedure Calls
+
 A *Dealer* or *Callee* canceled a call previously issued
 
 {align="left"}
--- a/rfc/text/advanced/ap_48_uris.md
+++ b/rfc/text/advanced/ap_48_uris.md
@@ -34,7 +34,7 @@ The authentication of the *Client* failed technically at run-time, and the *Clie
 {align="left"}
         wamp.error.authentication_failed
 
-The *Client* did not authenticate (at all) and a successful (non-anonymous) authentication is required.
+The *Client* did not provide the required, non-anonymous, authentication information.
 
 {align="left"}
         wamp.error.authentication_required
--- a/rfc/text/advanced/ap_48_uris.md
+++ b/rfc/text/advanced/ap_48_uris.md
@@ -29,7 +29,7 @@ The authentication as presented by the *Client* is denied (e.g. "wrong password"
 {align="left"}
         wamp.error.authentication_denied
 
-The authentication of the *Client* failed technically at run-time, and the *Client* is therefor rejected ("fail secure operation").
+The *Client* authentication was rejected due to a technical runtime failure ("fail secure" operation).
 
 {align="left"}
         wamp.error.authentication_failed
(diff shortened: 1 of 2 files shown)


rfc/text/advanced/ap_48_uris.md
```diff
@@ -2,6 +2,62 @@
...
 WAMP pre-defines the following error URIs for the **Advanced Profile**. WAMP peers SHOULD only use the defined error messages.
 
+## Authentication
+
+No authentication method the *Client* offered is accepted.
+
+{align="left"}
+        wamp.error.no_auth_method
+
+The *Client* authenticated for a *Realm* (`realm|string`) that does not (or no longer) exists.
```
KSDaemon, line 12: Mmm... I don't like the phrase `authenticated` because technically Client was not authenticated, the Client only tried to authenticate :)

So mb rephrase it to:
```
The *Client* tries to authenticate for a *Realm* (`realm|string`) that does not (or no longer) exists.
```
or 
```
The *Client* tried to authenticate for a *Realm* (`realm|string`) that does not (or no longer) exists.
```

Well... Of course, my English is not so good as @ecorm's :)

rfc/text/advanced/ap_48_uris.md
```diff
@@ -2,6 +2,62 @@
...
+        wamp.error.authentication_failed
+
+The *Client* did not authenticate (at all) and a successful (non-anonymous) authentication is required.
+
+{align="left"}
+        wamp.error.authentication_required
+
+## Authorization
+
+The *Principal* is not authorized to perform such *Action*.
```
KSDaemon, line 44: Here is an introduced kind of new term *Principal*. AFAIK we do not use it across the SPEC (can't find anywhere in the spec). Imo it is a pretty wide term. Why not keep the *Client*?

Btw, [Authentication Methods](https://wamp-proto.org/wamp_latest_ietf.html#name-authentication-methods) uses Client everywhere. So I think it is important to keep this all in sync. Because in this case reader is already familiar with the terms he/she saw in the spec and nothing surprising here. If we want to add new terms - they should be added and updated across all the relevant sections of the spec and also given some descriptions.

rfc/text/advanced/ap_48_uris.md
oberstet: agreed. I'll change that.

rfc/text/advanced/ap_48_uris.md
oberstet: rgd "Principal", that's right!

IMO, it is another gap. The historic reason is: we initially said in the spec "all authentication and authorization is implementation specific (outside the spec)", then we added authentication methods, but did not spec any authorization mechanisms.

- **A**uthentication: who is it?
- **A**uthorization: is it (the action by "who" on the respective "object") allowed?
- **A**uditability: what was actually done in the past (a log)

now that we want to have authorization related URIs in the spec, we a term for "who"

A WAMP *Session* is established originating from a *Client* to a *Router*, and when the *Session* performs *Authentication*, and that authentication is successful, the *Router* will have established

* *Realm*
* *AuthRole*
* *AuthId*

_for_ that *Session* running _in_ that *Client*.

The triple "(Realm, AuthRole, AuthId)" is a *Principal*.

-----

> Btw, [Authentication Methods](https://wamp-proto.org/wamp_latest_ietf.html#name-authentication-methods) uses Client everywhere

this section is about Authorization, not Authentication. this is the 2nd A in "AAA"

oberstet: I am inclined to remove the section https://github.com/wamp-proto/wamp-proto/blob/master/rfc/text/basic/bp_01_introduction.md#router-implementation-specifics completely.

what's not in the spec is not in the spec and obviously then implementation defined. why state a conclusive list when we're working on such things to add to the spec. it doesn't add much value IMO

KSDaemon: Yeah, I see.

My suggestion is to: 

1. Leave the term `Сlient` for this PR and merge it. It will be consistent with other chapters. This will push us a step forward.
2. Continue AAA refactoring in a separate issue.

KSDaemon: > I am inclined to remove the section https://github.com/wamp-proto/wamp-proto/blob/master/rfc/text/basic/bp_01_introduction.md#router-implementation-specifics completely.

Yeah 👍🏻  Why even try to mention things that lies out of the SPEC scope at all :)

rfc/text/advanced/ap_48_uris.md
```diff
@@ -2,6 +2,62 @@
...
+
+{align="left"}
+        wamp.error.no_such_principal
+
+The authentication as presented by the *Client* is denied (e.g. "wrong password").
+
+{align="left"}
+        wamp.error.authentication_denied
+
+The authentication of the *Client* failed technically at run-time, and the *Client* is therefor rejected ("fail secure operation").
```
ecorm, line 32:
```suggestion
The *Client* authentication was rejected due to a technical runtime failure ("fail secure" operation).
```

rfc/text/advanced/ap_48_uris.md
```diff
@@ -2,6 +2,62 @@
...
+
+{align="left"}
+        wamp.error.authentication_failed
+
+The *Client* did not authenticate (at all) and a successful (non-anonymous) authentication is required.
```
ecorm, line 37:
```suggestion
The *Client* did not provide the required, non-anonymous, authentication information.
```

ecorm: Referring to success here is confusing, because there are URI devoted to that: `wamp.error.authentication_denied` and `wamp.error.authentication_failed`.

rfc/text/advanced/ap_48_uris.md
```diff
@@ -2,6 +2,62 @@
...
+        wamp.error.authentication_required
+
+## Authorization
+
+The *Principal* is not authorized to perform such *Action*.
+
+{align="left"}
+        wamp.error.authorization_denied
+
+Authorization of the *Principal* for the *Action* could not be determined as it failed technically at run-time, and the *Action* is therefor rejected ("fail secure operation").
```
ecorm, line 49:
```suggestion
Authorization of the *Principal* to perform the given *Action* was rejected due to a technical runtime failure ("fail secure" operation).
```

rfc/text/advanced/ap_48_uris.md
```diff
@@ -2,6 +2,62 @@
...
+
+{align="left"}
+        wamp.error.authorization_failed
+
+Authorization of the *Principal* is required for each individual execution of the *Action*. This can be used for capability-based access control.
```
ecorm, line 54:
```suggestion
Authorization of the *Principal* is required to perform the given *Action*. This can be used for capability-based access control.
```

rfc/text/advanced/ap_48_uris.md
```diff
@@ -2,6 +2,62 @@
 
 WAMP pre-defines the following error URIs for the **Advanced Profile**. WAMP peers SHOULD only use the defined error messages.
 
+## Authentication
+
+No authentication method the *Client* offered is accepted.
+
+{align="left"}
+        wamp.error.no_auth_method
```
ecorm, line 10:
```suggestion
        wamp.error.no_matching_auth_method
```

ecorm (changes requested):
Finished review

rfc/text/advanced/ap_48_uris.md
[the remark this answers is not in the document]
oberstet: I agree. A proposal for a better URI: `wamp.error.no_acceptable_authentication_method`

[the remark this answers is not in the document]
oberstet: resolved by merge of your proposal below

rfc/text/advanced/ap_48_uris.md
```diff
@@ -2,6 +2,62 @@
...
+
+{align="left"}
+        wamp.error.no_matching_auth_method
+
+The *Client* attempted to authenticate for a non-existing *Realm* (`realm|string`).
+
+{align="left"}
+        wamp.error.no_such_realm
+
+The *Client* attempted to authenticate for a *Role* (`authrole|string`) that does not (or no longer) exists.
```
ecorm, line 17:
```suggestion
The *Client* attempted to authenticate for a non-existing *Role* (`authrole|string`).
```

rfc/text/advanced/ap_48_uris.md
```diff
@@ -2,6 +2,62 @@
...
+
+{align="left"}
+        wamp.error.no_such_role
+
+The *Client* authenticated for a *Principal* (`authid|string`) that does not (or no longer) exists.
```
ecorm, line 22:
```suggestion
The *Client* authenticated for a non-existing *Principal* (`authid|string`).
```

ecorm (changes requested):
A couple more suggestions.

KSDaemon (approved):
Great! There's no reason not to flush this PR! 👍🏻

SunFulong:
Sorry for missing this round, it seems more complete than previous discuss now! You are moving too fast!

PR #453, opened 2023-02-27
[WIP] Progressive call invocations

oberstet:
this is intended to consolidate https://github.com/wamp-proto/wamp-proto/pull/451, the discussion there, clean up a bunch, and actually propose a complete feature description including call cancellation and handling of race conditions


oberstet:
@KSDaemon ah, shit, this is now conflicting as I merged your PR:(

KSDaemon:
Hey folks! Let's choose the final term for `Progressive calls`! 

I see that @oberstet Already has changed to `Progressive call arguments`.

But I want to discuss it yet more.

I don't like the word `arguments` as it relates almost to every message with payload. Because every payload has positional and key-value arguments.

Even the phrase `Progressive call arguments` without context for not-familiar-with-wamp-spec reader may lead to confusion. Because even progressive call result has arguments! 

What do you have against naming it `Progressive call invocation`? 
In my view, it is way better. 

We have call invocations that they certainly say about call direction from Caller → to Callee.
And we have results — that also definitely say about message flow direction from Callee → to Caller.
Also, these terms contain WAMP message type in it and it is super easy for anyone to map them to underlying message flows without faults.

`Progressive call invocations` + `Progressive call results` = ❤️ `Progressive calls` (~peace~ ~friendship~)

Of course, If most people prefer another term, so be it.

What do you say, guys? @oberstet @Jopie64 @ecorm

ecorm:
> I don't like the word arguments

I was not fond of "Progressive Call Arguments" either and I prefer your suggestion of "Progressive Call Invocations", which is more symmetrical with "Progressive Call Results", having a message type name at the end for both.

oberstet:
> I don't like the word arguments as it relates almost to every message with payload.

right. I agree. that could be easily confusing.

> Progressive call invocations

this doesn't have the "arguments" problem ... the only problem is INVOCATION is a message on the Dealer-to-Callee leg only, while the progressive XX actually starts from the Caller

and at the same time "Progressive call results" has "results" in it, where RESULT is a message on the Dealer-to-Caller leg (the Callee produces YIELDs) - IOW: this then a asymmetry in naming vs above (Callee vs Caller is the reference point to derive the term)

BUT: there likely isn't a "perfect wording", and hence: I'd be fine with "Progressive call invocations"!

ecorm:
> Progressive call invocations + Progressive call results = Progressive calls

This would be confusing to those who still have the old definition of "Progressive Calls" stuck in their head.

The two are mostly described independently of each other, except for that recent subsection I added that talks about both. It would be better to just say the long form "Progressive Call Results and Progressive Call Invocations" in the very few places that mention the two.

oberstet:
> Just a note here that implementers 

what implementations are there? do you have one? with which client libraries was it tested? are there any examples?

fwiw, I am -1 on continuing to use "progressive Calls" for what is "progressive call arguments / invocations / XYZ", and also introducing synonyms. I care much less what "progressive call XYZ" ... just that it is different from only "progressive calls"

KSDaemon:
I agree that for now there are not so many implementations of progressive call invocations. I'm aware of Nexus implemented by me :))
And plans to implement it in some other under my wing :)

KSDaemon:
Well, then it's settled: shall we use the term `Progressive Call Invocations`?

ecorm:
> what implementations are there? do you have one? with which client libraries was it tested? are there any examples?

These are good questions. I had assumed that at least KSDaemon had one, seeing that he authored that feature in the first place. He just confirmed that he does indeed have a working implementation.

I was not suggesting that this alias be included in the spec, but that implementors should be aware about the possibility of being backwards-compatible with the old feature name.

Jopie64:
FWIW although I do see the asymetry @oberstet is talking about, I'm also in favor of "Progressive Call Invocations".

It has the added benefit that progressive invocations can be argument payload less :)

ecorm:
>FWIW although I do see the asymetry oberstet is talking about

It is symmetrical from the point of view of messages emitted by the Dealer:

```mermaid
flowchart LR
    Caller-->|CALL|Dealer
    Dealer-->|RESULT|Caller
    Dealer-->|INVOCATION|Callee
    Callee-->|YIELD|Dealer
```

rfc/STATUS.md
```diff
@@ -51,7 +51,7 @@ final          | The section does not require any more attention.
 |   Features                        |               |               | [advanced/ap_features.md](text/advanced/ap_features.md) |  |
 |   Adv. RPC Features               | -             |               | -
 |     Progressive Call Results      | beta          |               | [advanced/ap_rpc_progressive_call_results.md](text/advanced/ap_rpc_progressive_call_results.md) | https://github.com/wamp-proto/wamp-proto/issues/199 |
-|     Progressive Calls             | sketch        |               | [advanced/ap_rpc_progressive_calls.md](text/advanced/ap_rpc_progressive_calls.md) | https://github.com/wamp-proto/wamp-proto/issues/167 |
+|     Progressive Call Arguments    | alpha         |               | [advanced/ap_rpc_progressive_call_arguments.md](text/advanced/ap_rpc_progressive_call_arguments.md) | https://github.com/wamp-proto/wamp-proto/issues/167 |
```
KSDaemon, line 54: So we all agreed to rename this to `Progressive Call Invocations`

KSDaemon (changes requested):

oberstet:
awesome!! so we finally have consensus  on **Progressive Call Invocations** making everyone happy=)

> that means, this PR needs to be adjusted accordingly. I'd only be able to change it coming weekend since I'm busy this week onsite working at some customer - should anyone else wants to change / slash it before, np, pls got ahead, would be cool!

oberstet:
> He just confirmed that he does indeed have a working implementation.

did he? with what client libraries did he test? did he test one of those test programs using the feature against a 2nd router implementation that also claims support?

do we have any "rule" as rgd the "status" of the feature in https://github.com/wamp-proto/wamp-proto/blob/master/rfc/STATUS.md ?

do we want to require at least two router implementations with "working code", interoperable for one given client?

do we want to require at least two client implementations able to work with any of above two router implementations?

oberstet:
> And plans to implement it in some other under my wing :)

curious, what's under that wing? ;)

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
```diff
@@ -1,25 +1,25 @@
...
 
 * Payload is too big to send it whole in one request, e.g., uploading a file.
 * Long-term data transfer that needs to be consumed early, such as a media stream.
 * RPC is called too often and overall processing can be optimized: avoid the need to generate a new id for requests,
   initiate data structures for a new call, etc.
 
-In such cases, a procedure implemented by a *Callee* and registered at a *Dealer* may be made to receive progressive calls,
-where the *Callee* waits after receiving the entire set of payload chunks before sending the result. 
+In such cases, a procedure implemented by a *Callee* and registered at a *Dealer* may be made to receive progressive call invocations,
+where the *Callee* waits after receiving the entire set of payload chunks before sending the result.
```
KSDaemon, line 11: Looking at this sentence, I start thinking that we should rewrite it. As `waits after receiving the entire set of payload` is not the only possible way. And *Callee* may start processing data ASAP. And even return chunks of results (with the help of progressive call results)

KSDaemon:
Regarding working implementations:

Well, it is implemented in Nexus on the router and client sides :) So +1 router +client. Unit tests check that.

I don't want to name any deadlines or names, but we use many different wamp clients in different languages and stacks. I think that as the need arises (or my free time) this feature will appear in other implementations as well!

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
ecorm: We are describing the Progressive Call Invocations feature independently of Progressive Call Results at this point. The new *Progressive Calls with Progressive Call Results* subsection I consolidated discussed the mixing of both.

ecorm:
> I think that as the need arises (or my free time) this feature will appear in other implementations as well!

It will be in mine, but it is still WIP. I have adopted a streaming API for this feature, which hides this notion of RPCs that don't match the semantics of their programming language counterparts.

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
oberstet: what is a "payload chunk"? this term isn't defined or required

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
```diff
@@ -126,19 +126,19 @@ To bind `INVOCATION` messages to the same *Callee*, the *Caller* can specify the
         CALL.Options.sticky|bool := true
 
 
-In this case, the *Dealer* must make a first `INVOCATION` based on `<invocation_policy>` and then route all 
+In this case, the *Dealer* must make a first `INVOCATION` based on `<invocation_policy>` and then route all
 subsequent `progressive` calls to the same *Callee*.
 
-If binding all ongoing `progressive` calls to the same *Callee* is not required, the *Caller* can set the `sticky` option to `false`. 
+If binding all ongoing `progressive` calls to the same *Callee* is not required, the *Caller* can set the `sticky` option to `false`.
```
Jopie64, line 132: Hmmm this part was removed right? Or is it because this change is based on an older version of the docs?

KSDaemon: kinda Yes and No :)
The original PR with removal was merged but then due to conflicts was reverted. So we have an opened #460 once again. But it needs to be merged after this one. And then — everything should be all good %)

KSDaemon: FYI @Jopie64 Merged PR with sticky :)

oberstet:
TODO: add reverted changes from https://github.com/wamp-proto/wamp-proto/pull/466 into this PR

ecorm:
While you're at it, please remove the checkmark under `Cr` for Call Timeout feature in the [Feature Announcement](https://wamp-proto.org/wamp_latest_ietf.html#name-feature-announcement) table. I missed it in my recent Call Timeout PR.

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
KSDaemon, line 11:
```suggestion
where the *Callee* may start processing the incoming data without awaiting the entire set of payload chunks.
```

KSDaemon: @oberstet @ecorm What do you think of this change?

KSDaemon (commented):
Almost LGTM! 

Just one sentense need agreement that I mentioned

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
oberstet: it seems the above 2 paragraphs are to describe the 2 options a callee that receives progressive call invocations (not "payload chunks") does have. start processing _and_ start producing (progressive) results immediately OR start receving progressive invocations and processing (or not) immediately, but continue to do so before finally producing 1 non-progressive results.

if so, I agree: it makes sense to explicitly describe those differing behaviors possible for a callee

KSDaemon: agree! 👍🏻

ecorm: > What do you think of this change?

LGTM

ecorm (changes requested):
Review done

ecorm:
There needs to be an addition that the Dealer must support call cancellation. This follows from this requirement that Callees support call cancellation:

> Progressive Calls can work only if all three peers support and announce this feature. In addition, *Callees* MUST announce support of the Call Cancelling feature via
>
>    `HELLO.Details.roles.callee.features.call_cancelling|bool := true`

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
```diff
@@ -28,11 +28,11 @@ In addition, *Callees* MUST announce support of the _Call Cancelling_ feature vi
 The following cases, where a *Caller* sends a `CALL` message with `progress := true`, MUST be treated as *protocol errors*
```
ecorm, line 28:
```suggestion

As a consequence, *Dealers* MUST also announce support of the _Call Cancelling_ feature via

{align="left"}
        WELCOME.Details.roles.dealer.features.call_cancelling|bool := true

The following cases, where a *Caller* sends a `CALL` message with `progress := true`, MUST be treated as *protocol errors*
```

ecorm (changes requested):
Dealer cancellation feature announcement needed for Progressive Call Invocations

oberstet:
> There needs to be an addition that the Dealer must support call cancellation. This follows from this requirement that Callees support call cancellation:

why would that be _required_? in any case, support for any of progressive call invocations or call results seems independent of support for call cancellation

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
oberstet: -1

ecorm:
> why would that be required? in any case, support for any of progressive call invocations or call results seems independent of support for call cancellation

Because it's required in the spec??  See https://wamp-proto.org/wamp_latest_ietf.html#section-11.2-31

> If the original *Caller* is no longer available for some reason (has left the realm), then the *Dealer* shall send an INTERRUPT to the *Callee.*

oberstet:
> Because it's required in the spec??

thanks for reminding me this already has been agreed and put into the spec;)

> it makes sense in a way to demand cancel if progress is supported - even though it isn't logically required strictly

KSDaemon (approved):
Hey folks! Seems that we need just a one more step to land this! 
Maybe we can do it? :)

ecorm:
> Hey folks! Seems that we need just a one more step to land this!

What is there left to do?

KSDaemon:
> What is there left to do?

Merge! :) At least me — approved. So if there are no objections...

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
```diff
@@ -1,38 +1,44 @@
-## Progressive Calls {#rpc-progressive-calls}
+## Progressive Call Invocations {#rpc-progressive-call-invocations}
 
-A *Caller* may issue a progressive call. This can be useful in a few cases:
+A *Caller* may issue a progressive call having progressive invocations. This can be useful in a few cases:
```
ecorm, line 3:
```suggestion
A *Caller* may issue a call having progressive invocations. This can be useful in a few cases:
```

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
```diff
@@ -112,33 +118,22 @@ If the *Callee* supports progressive calls, the *Dealer* shall forward the *Call
...
-In this case, progressive `CALL` messages can be routed to different *Callees*, which can lead to unexpected results.
-To bind `INVOCATION` messages to the same *Callee*, the *Caller* can specify the `sticky` option during `CALL`
-
-{align="left"}
-        CALL.Options.sticky|bool := true
-
-
-In this case, the *Dealer* must make a first `INVOCATION` based on `<invocation_policy>` and then route all 
+However, allowing progressive `CALL` messages to be routed to different *Callees* would lead to unexpected results.
+To prevent this the *Dealer* must make a first `INVOCATION` based on `<invocation_policy>` and then route all
```
ecorm, line 129:
```suggestion
To prevent this the *Dealer* MUST make a first `INVOCATION` based on `<invocation_policy>` and then route all
```

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
```diff
@@ -180,28 +175,26 @@ The message flow for cancelling progressive calls involves:
...
-**Progressive Calls with Progressive Call Results**
+*Progressive Call Invocations* may be used in conjunction with *Progressive Call Results* if the *Caller*, *Dealer*, and *Callee* all support both features. This allows the *Callee* to start sending partial results back to the *Caller* after receiving one or more initial payload chunks. Efficient two-way streams between a *Caller* and *Callee* can be implemented this way.
 
-Progressive calls may be used in conjunction with *Progressive Call Results* if the *Caller*, *Dealer*, and *Callee* all
-support both features. This allows the *Callee* to start sending partial results back to the *Caller* after receiving
-one or more initial payload chunks. Efficient two-way streams between a *Caller* and *Callee* can be implemented this way.
-  
-The following message flow illustrates a progressive call when a *Callee* starts sending progressive call 
-results immediately. Note that `YIELD` messages don't need to be matched with `CALL`/`INVOCATION` messages. 
+The following message flow illustrates a progressive call using progressive call invocations when a *Callee* starts sending progressive call
```
ecorm, line 196:
```suggestion
The following message flow illustrates a call using progressive call invocations when a *Callee* starts sending progressive call
```

ecorm (changes requested):
I did a final read and have a few very minor nitpicks.

PR #454, opened 2023-02-27
Terminology chapter

oberstet:


rfc/text/wamp-terminology.md
```diff
@@ -0,0 +1,74 @@
...
+|-----------------------------------|-------------------------------------------------------------------------------------------------------|
+| *User*                            | A user runs *Client*s or *Router*s which implement WAMP                                               |
+| *Client*                          | A program run by a *User* with application code using WAMP for application level communication        |
+| *Router*                          | A program run by a *User* with middleware code using WAMP to provide application routing services     |
+| *Peer*                            | A WAMP *Client* or *Router*. An implementation might embed and use both roles                         |
+| *Realm*                           | Isolated WAMP URI namespace, routing and administrative domain, optionally protected by **AA**        |
+| *Transport*                       | The bidirectional, ordered, full duplex message channel over which *Peers* communicate                |
+| *Connection*                      | When a *Client* is using a *Transport* over a network, the underlying network connection              |
```
KSDaemon, line 28: The connection can be made locally, via unix socket, interprocess RPC, or even internally inside one running program. So I think it should be not bound to the network but rather focus on the logical communication channel between peers.

rfc/text/wamp-terminology.md
```diff
@@ -0,0 +1,74 @@
...
+
+{align="left"}
+| Term                              | Definition                                                                                            |
+|-----------------------------------|-------------------------------------------------------------------------------------------------------|
+| *Authentication*                  | Establishes the identity of a *Session* on a *Realm*                                                  |
+| *Principal*                       | Once authenticated, *Session*s identify under a *Principal*                                           |
+| *Credentials*                     | Any authentication information and secrets used during *Authentication*                               |
+| *Authorization*                   | Decides about permission for a given *Action* on an URI or URI pattern by a *Principal*               |
+| *Access Control*                  | Policy for selective restriction of access by *Action*s on URIs or URI patterns and *Principal*s      |
+| *Role-based Access Control*       | An *Access Control* policy based on *Realm* (`realm`) and *Role* (`authrole`) of a *Principal*        |
```
KSDaemon, line 42: I would add also a short well-known version: *Role-based Access Control* (*RBAC*)

KSDaemon (commented):
Looks very promissing :)
Just a few notes!

rfc/text/wamp-terminology.md
```diff
@@ -0,0 +1,80 @@
...
+| *Principal*                        | Once authenticated, *Session*s identify under a *Principal*                                            |
+| *Credentials*                      | Any authentication information and secrets used during *Authentication*                                |
+| *Authorization*                    | Decides about permission for a given *Action* on an URI or URI pattern by a *Principal*                |
+| *Access Control*                   | Policy for selective restriction of access by *Action*s on URIs or URI patterns and *Principal*s       |
+| *Role-based Access Control (RBAC)* | An *Access Control* policy based on *Realm* (`realm`), *Role* (`authrole`) of *Principal* and *Action* |
+| *Discretionary Access Control*     | *FIXME*                                                                                                |
+| *Mandatory Access Control*         | *FIXME*                                                                                                |
+| *Capability-based Access Control*  | *FIXME*                                                                                                |
```
KSDaemon, line 51: Well, if gonna mentioning different possible ACLs then this list is missing Attribute-based access control (ABAC)

But IDK, do we really need to describe this? I think WAMP as a protocol provides the building blocks on top of which any kind of security scheme may be built. We do not restrict or encourage or even suggest using specific security schemes. So mb it is better not to mention it at all?

rfc/text/wamp-terminology.md
```diff
@@ -0,0 +1,80 @@
...
+|------------------------------------|--------------------------------------------------------------------------------------------------------|
+| *Authentication*                   | Establishes the identity of a *Session* on a *Realm*                                                   |
+| *Principal*                        | Once authenticated, *Session*s identify under a *Principal*                                            |
+| *Credentials*                      | Any authentication information and secrets used during *Authentication*                                |
+| *Authorization*                    | Decides about permission for a given *Action* on an URI or URI pattern by a *Principal*                |
+| *Access Control*                   | Policy for selective restriction of access by *Action*s on URIs or URI patterns and *Principal*s       |
+| *Role-based Access Control (RBAC)* | An *Access Control* policy based on *Realm* (`realm`), *Role* (`authrole`) of *Principal* and *Action* |
+| *Discretionary Access Control*     | *FIXME*                                                                                                |
```
KSDaemon, line 49: The definition for this in the context of WAMP may be smth like this:

An *Access Control* policy based on lists or matrices of URIs and *Actions* mapped to *Principal*s

rfc/text/wamp-terminology.md
```diff
@@ -0,0 +1,80 @@
...
+| *Mandatory Access Control*         | *FIXME*                                                                                                |
```
KSDaemon, line 50: I think MACs in case of WAMP may be based on Trust Levels for the *Principals*

rfc/text/wamp-terminology.md
```diff
@@ -0,0 +1,80 @@
...
+| *Dealer*                          | *FIXME*                                                                                                 |
+| *Peer*                            | A WAMP *Client* or *Router*. An implementation might embed, provide or use both roles                   |
+| *Realm*                           | Isolated WAMP URI namespace, routing and administrative domain, optionally protected by **AA**          |
+| *Transport*                       | A message-based, reliable, ordered, bidirectional (full-duplex) channel over which *Peers* communicate  |
+| *Connection*                      | An underlying entity (if any) carrying the *Transport*, e.g. a network connection, pipe, queue or such  |
+| *Session*                         | Transient conversation between a *Client* and a *Router* on a *Realm* over a *Transport*                |
+| *Message*                         | *FIXME*                                                                                                 |
+| *Serializer*                      | *FIXME*                                                                                                 |
```
KSDaemon, line 33: Interesting how to describe a serializer without saying that: `Serializer is just a serializer` :)

How about this: 

The Serializer is responsible for turning data objects and WAMP messages into a specific format convenient for transporting (JSON, CBOR, MsgPack, etc) and vice versa.

rfc/text/wamp-terminology.md
```diff
@@ -0,0 +1,80 @@
...
+| *Broker*                          | *FIXME*                                                                                                 |
+| *Dealer*                          | *FIXME*                                                                                                 |
+| *Peer*                            | A WAMP *Client* or *Router*. An implementation might embed, provide or use both roles                   |
+| *Realm*                           | Isolated WAMP URI namespace, routing and administrative domain, optionally protected by **AA**          |
+| *Transport*                       | A message-based, reliable, ordered, bidirectional (full-duplex) channel over which *Peers* communicate  |
+| *Connection*                      | An underlying entity (if any) carrying the *Transport*, e.g. a network connection, pipe, queue or such  |
+| *Session*                         | Transient conversation between a *Client* and a *Router* on a *Realm* over a *Transport*                |
+| *Message*                         | *FIXME*                                                                                                 |
```
KSDaemon, line 32: I'm thinking about smth like this:

The minimal indivisible unit of information transmitted between peers.

rfc/text/wamp-terminology.md
oberstet: yeah, I see what you mean. on the other hand, my thinking of adding call these terms was to give the reader more clues of how stuff would fit together ... "terms systematics". fwiw, crossbar currently only implements RBAC, which is one flavor of a MAC. and I'd only be interested in - maybe - capabilities based scheme, but if so, likely only in the new router implementation I'll start ...

we should IMO have all terms that are used somewhere else in the spec text, even if only as a hint.

eg CAP is mentioned somewhere even though there is no router yet ...

anyways, I'm fine with all options: remove it for now, or I can add text coming weekend, or ...

rfc/text/wamp-terminology.md
oberstet: How do you look at the "difference" between ACLs and AC-matrices?

I'm thinking like:

ACL: `(URI, Action) -> [Principal]`
ACM: `{Principal X Action} -> Actions`

rfc/text/wamp-terminology.md
oberstet: A Serializer is a function from WAMP Messages to bytes - as bytes (a chunk of) is what is needed by WAMP Transports.

So Serializer, Message, Transport and Bytes (of a message as taken/consumed/transported by Transport) all relate

rfc/text/wamp-terminology.md
oberstet: good Q ..  what _is_ a WAMP Message.

" minimal indivisible unit of information transmitted between peers."

I like that!  maybe "exchanged" instead of "transmitted"

rfc/text/wamp-terminology.md
oberstet: actually, there are two meanings of "serializer" in WAMP. let me explain.

technically and in general (independent of WAMP), a "raw serializer" **RS** is

1. if you take **O** as the set of all possible values of objects in the host language, and
2.  **B** the set of all Bytes strings which encode possible values which are valid under **RS**

then **RS** is a function `RS: O -> B`.

couple of notes:
* of course no every Bytes string is "proper" under **RS** (there are Bytes string which are invalid, as hence **B** is a subset of the set of all Bytes string)
* of course **RS** should be "universal" with respect to the host language (every valid host language value should be serializable under **RS** .. otherwise some application payloads might not be possible to be sent or received in WAMP actions)

Moreover, what's interesting to note:

* it would be desirable if **RS** defined a *bijective* function  `RS: O -> B`
* however, the only serializer with that property I know of is [Bencode](https://en.wikipedia.org/wiki/Bencode)
* none of the currently defined WAMP serializers has this property

now, given **RS**, when specifically encoding WAMP messages with arbitrary application payloads. the application payloads are arbitrary (if the serializer is universal), but the fact that this is about WAMP messages means, the "outer" stuff is restricted by the WAMP spec (for WAMP messages). let's call that serializer **WS**. then

1. the set of all valid WAMP messages **W**, that is objects in the host language which are valid WAMP message with any application payload, and have arbitrary application payloads
2. the set of all Bytes strings **BW** which **WS** deserializes into messages in **W** with arbitrary application payloads, then

**WS** is a function `WS: W->BW`

rfc/text/wamp-terminology.md
KSDaemon: `The minimal indivisible unit of information exchanged between peers` sounds good! 👍🏻

ecorm: > however, the only serializer with that property I know of is [Bencode](https://en.wikipedia.org/wiki/Bencode)

I'm not sure, but [Deterministically-Encoded CBOR](https://www.rfc-editor.org/rfc/rfc8949.html#name-deterministically-encoded-c) may also have that property. Two recent implementations of such have been [announced on the CBOR mailing list](https://mailarchive.ietf.org/arch/msg/cbor/l7nzQHFjfpK9nfBOHiQ1L-Rr558/).

rfc/text/wamp-terminology.md
ecorm: > however, the only serializer with that property I know of is [Bencode](https://en.wikipedia.org/wiki/Bencode)

I'm not sure, but [Deterministically-Encoded CBOR](https://www.rfc-editor.org/rfc/rfc8949.html#name-deterministically-encoded-c) may also have that property. Two recent implementations of such have been [announced on the CBOR mailing list](https://mailarchive.ietf.org/arch/msg/cbor/l7nzQHFjfpK9nfBOHiQ1L-Rr558/).

KSDaemon: Well, that is cool and maybe even interesting to some readers. But let's start with simple descriptions and then we can elaborate more on that.

rfc/text/wamp-terminology.md
KSDaemon: Yeah, I see ACL mostly the same. Also, ACLs may be expressed from the other side: `{Principal} -> [URI + Action]`

I see ACM more like this: `{Principal X URI} -> [Action]`

| | Uri 1 | Uri 2 |
| --- | --- | --- |
| Principal 1 | publish | call |
| Principal 2 | publish+subscribe | call+register |


But I think that all ways are eligible and as usual `it depends` :)

rfc/text/wamp-terminology.md
KSDaemon: > we should IMO have all terms that are used somewhere else in the spec text, even if only as a hint.

Agree. That is good.

rfc/text/wamp-terminology.md
ecorm: For "serializer", how about: "Pair of functions that convert a Message (or Message Payload) into bytes and vice-versa"

In the spec, the term "serializer" is used loosely to also include deserialization operations. In electronics, [SerDes](https://en.wikipedia.org/wiki/SerDes), refers to a block that performs both serialization and deserialization, but it is not commonly used in software.

_Codec_ (coder/decoder) is a term that is more commonly found in software, but most people think "audio/video" when they see that term.

I think most readers will understand from context that "serializer" also includes the facilities for converting bytes back into messages/payloads.

rfc/text/wamp-terminology.md
ecorm: "minimal indivisible" is technically redundant; you could just say "indivisible". :grin:

ecorm (changes requested):
Finished review

rfc/text/wamp-terminology.md
```diff
@@ -0,0 +1,76 @@
...
+| *Serializer*                      | A *Serializer* encodes WAMP messages with application payloads into bytes strings for transport         |
+
+
+### Authentication and Authorization (AA)
+
+{align="left"}
+| Term                               | Definition                                                                                                                     |
+|------------------------------------|--------------------------------------------------------------------------------------------------------------------------------|
+| *Authentication*                   | Establishes the identity of a *Session* on a *Realm*                                                                           |
+| *Principal*                        | Once authenticated, *Session*s identify under a *Principal*                                                                    |
```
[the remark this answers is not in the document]
oberstet: how about: "A user that can be authenticated on a realm, under a role"

https://en.wikipedia.org/wiki/Principal_(computer_security)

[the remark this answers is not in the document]
oberstet: technically, the principal is the `authid`  - and that is defined/relative to the realm.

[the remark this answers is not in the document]
ecorm: I think it's a has-a relationship between Principal and `authid`: A Principal **has** an authid.

Your suggestion of "A user that can be authenticated on a realm, under a role" seems fine, exception that I would substitute "user" with "entity", because *User* was previously defined as something else.

rfc/text/wamp-terminology.md
```diff
@@ -0,0 +1,75 @@
...
+|------------------------------------|--------------------------------------------------------------------------------------------------------------------------------|
+| *Authentication*                   | Establishes the identity of a *Session* within a *Realm*                                                                           |
+| *Principal*                        | The authenticated entity associated with a *Session* |
+| *Credentials*                      | The authentication information and secrets used during *Authentication*                                                        |
+| *Authorization*                    | A decision on permitting a *Principal* to perform a given *Action* on an *URI or URI pattern* |
+| *Access Control*                   | Policy for selective restriction of *Action*s on *URIs or URI patterns* performed by *Principal*s                             |
+| *Role-based Access Control (RBAC)* | An *Access Control* policy based on *Realm* (`realm`), *Principal*'s *Role* (`authrole`), *URI or URI pattern*, and *Action*   |
```
KSDaemon, line 43: If you don't mind, let's add also ABAC for the whole picture.

Attribute-based access control (ABAC) — Policy-based access control that defines an access control paradigm whereby a Principal's authorization to perform a set of operations is determined by evaluating attributes associated with the Principal, URI, Action, and, in some cases, environment attributes.

rfc/text/wamp-terminology.md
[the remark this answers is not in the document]
oberstet: mmh .. not sure ... my view:

> I think it's a has-a relationship between Principal and authid: A Principal has an authid.

No, the authid is just a string identifier for the principal - it is 1:1

The user-principal relation is 1:n, since one user can have different principals on realms

"User" is defined as "A person (or organization) running a WAMP *Client* or *Router*"

"Entity": is undefined and arbitrary

KSDaemon (commented):
Looks good. If you don't mind to add ABAC policy — let's make it and merge!

rfc/text/wamp-terminology.md
oberstet: I do mind;) I fact:

"Unlike [role-based access control (RBAC)](https://en.wikipedia.org/wiki/RBAC), which defines roles that carry a specific set of privileges associated with them and to which subjects are assigned, ABAC can express complex rule sets that can evaluate many different attributes."

fwiw, I deliberately decided against that and base policies only on authrole on realm - this includes APIs for dynamic auuthentication/authorization.

IOW: -1 from my side

KSDaemon: Well, maybe it makes sense. 

But just one simple example: let's say your dynamic authentication/authorization component takes into account in addition to authid/authrole time of the day to permit actions only during working hours. This is ABAC and not RBAC.

rfc/text/wamp-terminology.md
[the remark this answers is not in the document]
ecorm: > No, the authid is just a string identifier for the principal - it is 1:1

What I meant to say is that `authid` is an attribute of a Principal (technically, a has-a relationship with a String). That's not the same as saying the principal is the authid.

This class diagram models the way I understand it:
```mermaid
classDiagram
    direction LR
    User "1" -- "*" Principal
    Principal "*" o-- "1" Role
    class User {
        <<actor>>
    }
    class Principal {
        +authId: String
        +authRole: String
    }
```

Some apps will associate multiple roles to the same Principal, but WAMP is currently limited to a single role. I'm aware this has been brought up already in an issue.

From [Microsoft](https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-principals):
> A security principal is any entity that can be authenticated by the operating system, such as a user account, a computer account, or a thread or process that runs in the security context of a user or computer account, or the security groups for these accounts.

In the case of WAMP, if we apply this definition, a security principal can be a user account or a role (security group).

Using Microsoft's definition, the class diagram should really be:
```mermaid
classDiagram
    direction LR
    Principal <|-- Account
    Principal <|-- Role
    User "1" -- "*" Account
    Account "*" o-- "1" Role
    class User {
        <<actor>>
    }
    class Account {
        +authId: String
        +authRole: String
    }
```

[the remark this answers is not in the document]
ecorm: A WAMP Role can't be authenticated, so perhaps it should not be considered a Principal according to MS's definition. I don't know how one "authenticates" a security group. It's usually the account that's authenticated, and an account is assigned to one or more security groups. One could **authorize** a security group to perform an action, though.

rfc/text/wamp-terminology.md
oberstet: yes, that would be ABAC. and RBAC is a simple subset of ABAC. the name ABAC is slightly misleading IMO, as it really is much more than attributes of the principal, but

> subject's authorization to perform a set of operations is determined by evaluating attributes associated with the subject, object, requested operations, and, in some cases, environment attributes.

eg your example: the working hours is an attribute of the environment.

the complexity introduced is IMO unjustified for the gain, eg caching becomes pretty unclear how to do it safely

oberstet:
@KSDaemon @ecorm should we merge this one (we can always add more terms later)?

rfc/text/wamp-terminology.md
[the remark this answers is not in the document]
oberstet: No, I disagree: a Principal for a User only is defined in a Realm - it doesn't exist independently from a Realm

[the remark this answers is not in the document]
oberstet: > What I meant to say is that authid is an attribute of a Principal (technically, a has-a relationship with a String). That's not the same as saying the principal is the authid.

Ok, that makes sense. "has a". But the relation is only defined on a Realm. So:

User 1 --- Realm --- 1|0 Principal (authid)

A User can exist without a Principal (== authid) on a Realm.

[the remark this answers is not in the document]
oberstet: > A WAMP Role can't be authenticated, so perhaps it should not be considered a Principal 

Who said a Role would be considered a Principal?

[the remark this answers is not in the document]
oberstet: > One could authorize a security group to perform an action, though.

well, that's exactly how it is defined: the Role on a Realm, the URI and the Action determine the Access Controls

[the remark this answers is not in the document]
ecorm: > Who said a Role would be considered a Principal?

The Microsoft definition considers "security groups" to be a kind of Security Principal, and I think a Role could be considered a security group.

If you don't agree with MS's inclusion of "security groups" a kind of Security Principal, that's fine, as long as we define "Principal" in clear manner and be consistent in its usage.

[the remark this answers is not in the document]
ecorm: How about this?
Principal: A user account associated with a *Role*, having a string identifier (`authid`), that can be authenticated under a *Realm*

ecorm:
> Should we merge this one (we can always add more terms later)?

I'd like to nail down the definition of Principal. I made a suggestion upthread that I hope will please everybody:

Principal: A user account associated with a *Role*, having a string identifier (`authid`), that can be authenticated under a *Realm*

rfc/text/wamp-terminology.md
```diff
@@ -0,0 +1,75 @@
...
+both formally (open protocol and open source) as well as practically (switching costs).
+
+
+### Fundamental
+
+{align="left"}
+| Term                              | Definition                                                                                              |
+|-----------------------------------|---------------------------------------------------------------------------------------------------------|
+| *User*                            | A person (or organization) running a WAMP *Client* or *Router* |
+| *Client*                          | A program run by a *User*, with application code using WAMP for application level communications        |
```
ecorm, line 22:
```suggestion
| *Client*                          | A program run by a *User*, with application code using WAMP for application-level communications       |
```

rfc/text/wamp-terminology.md
```diff
@@ -0,0 +1,75 @@
...
+### Publish and Subscribe
+
+{align="left"}
+| Term             | Definition                                                                                                                    |
+|------------------|-------------------------------------------------------------------------------------------------------------------------------|
+| *Publisher*      | A *Publisher* is a *Session* that **publishes** application payloads to a (fully qualified) *Topic* for event routing         |
+| *Subscriber*     | A *Subscriber* is a *Session* that **subscribes** to a *Topic* to receive application payloads on matching events             |
+| *Topic*          | A *Topic* is an URI or URI pattern that can be subscribed to for event routing by *Subscriber*s                               |
+| *Subscription*   | A *Router* record resulting from a *Subscriber* successfully **subscribing** to a *Topic* for event routing      |
+| *Publication*    | A transient Router record resulting from a *Publisher* successfully **publishing** to a *Topic* for event routing         |
```
ecorm, line 74:
```suggestion
| *Publication*    | A transient *Router* record resulting from a *Publisher* successfully **publishing** to a *Topic* for event routing         |
```

rfc/text/wamp-terminology.md
```diff
@@ -0,0 +1,75 @@
...
+| *Event*          | A publication that is routed to a *Subscribers* having matching *Subscriptions* to the published *Topic*.              |
```
ecorm, line 75:
```suggestion
| *Event*          | A publication that is routed to *Subscribers* having matching *Subscriptions* to the published *Topic*.              |
```

ecorm (changes requested):
Minor grammar and formatting corrections. Please note that many of the `|`s at the end of table rows are no longer aligned due to the suggested changes. It was not possible for me to fix the `|` alignment via the GitHub review interface.

rfc/text/wamp-terminology.md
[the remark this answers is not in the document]
oberstet: I think the MS concept is mixing two concepts: Principals and Roles (what they call Groups). If a Principal has always one Role, and a Role is a Principal, this is mixing 2 different kinds of concepts. I am -1 on this.

----

> as long as we define "Principal" in clear manner and be consistent in its usage

following roughly the MS definition:

**Principal**: A security principal (`authid`) is any *User* that can be authenticated on a *Realm* (`realm`) that runs in the security context of a *Role* (`authrole`) on the *Realm*.

ecorm:
I just realized that these long tables might not be rendered properly in RFC format. Perhaps a simpler format should be used like:

```
Term: Definition goes here and can flow across several lines.

Another Term: Blank line between definitions helps to visually distinguish them.
```

oberstet:
> I just realized that these long tables might not be rendered properly in RFC format. Perhaps a simpler format should be used like:
> 
> ```
> Term: Definition goes here and can flow across several lines.
> 
> Another Term: Blank line between definitions helps to visually distinguish them.
> ```

maybe. too much work though. I think I'm done for now working on this PR, sorry:(

rfc/text/wamp-terminology.md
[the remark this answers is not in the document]
ecorm: Alright, I'll meet you halfway and agree to your latest definition with these grammatical changes:

Principal: A security principal (authid) is any User that can be authenticated under a Realm (realm) that runs in the security context of a Role (authrole) within that Realm.

"authenticate **under** a Realm" sounds better to my ear in the same way as "file that **under** receipts".

Perhaps the preposition "on" makes sense in German, but its doesn't sound right in English for this particular instance (neither in French, for that matter).

ecorm:
> maybe. too much work though. I think I'm done for now working on this PR, sorry:(

That's fine, we can save that tedium for another day. :grin:

rfc/text/wamp-terminology.md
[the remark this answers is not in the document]
oberstet: great! let's go with your last version.

fwiw, yeah, that wrong preposition is likely due to my native german tongue;) here is how it would read:

"Prinzipal: Ein Sicherheitsprinzipal ist jeder Nutzer der für einen Verwaltungsbereich authentifiziert ist und welcher unter dem Sicherheitskontext einer Rolle in diesem Verwaltungsbereich agiert."

this is definitely not entry level language. but anyways, it made be recognize 1 tiny more grammar glitch:

> Principal: A security principal (authid) is any User that can be authenticated under a Realm (realm) that runs in the security context of a Role (authrole) within that Realm.

the 2nd "that" could be read as to relate to Realm - where it should relate to User. ?? Maybe we should replace 2nd "that" with "and"? (I did that in the german version btw)

anyways, call me crazy, but thank you very much for caring about these details! I do as well. in particular in a terminology chapter.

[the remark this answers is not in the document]
oberstet: updated: https://github.com/wamp-proto/wamp-proto/pull/454/commits/157eb8c90a09bda54209eb0fb793470025246b92#diff-270c0c150b78b0712506f081bd98b0e462353bcaff0cdf77e2061b9cbe244989R39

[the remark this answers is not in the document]
ecorm: I see what you mean about the possible confusion, and I like your version better in your update.

oberstet:
alright, I merged this PR, hope everyone is OK for now, we can add more terms or adjust in follow up PRs as we see fit

PR #457, opened 2023-02-27
Fix typo in progressive call results doc

Jopie64:
Encountered typo while reading part of the docs.


oberstet (approved):
LGMT, also whitespace removed +1. merging (no substance matter change)

PR #469, opened 2023-03-12
Add CBON, UBJSON and FlatBuffers to RawSocket

SunFulong:
Adding CBOR to RawSocket like https://wamp-proto.org/wamp_latest_ietf.html#section-2.2-6 and https://wamp-proto.org/wamp_latest_ietf.html#section-2.3.1-4.

### DIFF
--- a/rfc/text/advanced/ap_38_transport_rawsocket.md
+++ b/rfc/text/advanced/ap_38_transport_rawsocket.md
@@ -76,7 +76,9 @@ The possible values for `SERIALIZER` are:
         1: JSON
         2: MessagePack
         3: CBOR
-        4 - 15: reserved for future serializers
+        4: UBJSON
+        5: FlatBuffers
+        6 - 15: reserved for future serializers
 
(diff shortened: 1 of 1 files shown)


KSDaemon:
@SunFulong Great! but that is only the beginning :) The whole Raw socket chapter (https://wamp-proto.org/wamp_latest_ietf.html#section-15.1) should be reviewed and updated accordingly ;)

oberstet:
fwiw, autobahn defines the following rawsocket serializer IDs

* `json`: `1`
* `msgpack`: `2`
* `cbor`: `3`
* `ubjson`: `4`
* `flatbuffers`: `5`

https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/serializer.py

KSDaemon:
So I think we should put those IDs into the spec.

oberstet (approved):

KSDaemon (approved):

Issue #477, opened 2023-04-29
Unclear behavior for caller leaving during CALL request


ecorm:
This is related to #476.

For **regular**, non-progressive calls, it's unclear how the router should react to the caller leaving in the middle of a CALL request, before the router receives a response from the callee.

Should the dealer send an INTERRUPT to the Callee if it supports call cancellation?

```mermaid
sequenceDiagram
    participant Caller
    participant Dealer
    participant Callee

    Caller->>Dealer: CALL
    Dealer->>Callee: INVOCATION
    Caller->>Dealer: GOODBYE
    Note right of Dealer: Should an INTERRUPT be sent to the  <br/>Callee if it supports call cancellation?
    Dealer->>Callee: INTERRUPT
```

The spec already mandates that the Dealer should send an INTERRUPT [in the case of progressive call results](https://wamp-proto.org/wamp_latest_ietf.html#section-11.1-55):

> Upon receiving a YIELD message from a _Callee_ with `YIELD.Options.progress == true` (for a call that is still ongoing), if the original _Caller_ is no longer available (has left the realm), then the _Dealer_ will send an INTERRUPT to the _Callee_. The INTERRUPT will have `Options.mode` set to `"killnowait"` to indicate to the client that no response should be sent to the INTERRUPT. This INTERRUPT is only sent in response to a progressive YIELD (`Details.progress == true`), **and is not sent in response to a normal or final YIELD**.

The above passage in bold is confusing, as INTERRUPTs are sent in response to a CANCEL, not YIELD. It also does not clearly specify what happens if the Dealer has not received any YIELD response yet.

For progressive call invocations, [the spec is much clearer](https://wamp-proto.org/wamp_latest_ietf.html#section-11.2-29):

> If the original _Caller_ is no longer available for some reason (has left the realm), then the _Dealer_ shall send an INTERRUPT to the _Callee_. The INTERRUPT MUST have `Options.mode` set to `"killnowait"` to indicate to the client that no response should be sent to the INTERRUPT.

I propose that behavior prescribed in progressive call invocations be applicable for _any_ type of CALL if the Callee supports call cancellation. For regular calls, if the Callee does not support cancellation, then the dealer simply discards the YIELD sent by the callee.

KSDaemon:
[...]

PR #478, opened 2023-05-01
Consolidate/clarify caller/callee leaving behavior

### LINKED ISSUE #476: Unspecified Behavior for Callee Leaving During CALL
This is a continuation of the long discussion in #448, but will focus purely on the behavior of a Callee leaving during a CALL request after the Dealer already sent the INVOCATION (either normal or progressive). The behavior for a Caller leaving during a normal call is underspecified as well, but I prefer to discuss that in a separate issue. I've learned the hard way that that's is easier to introduce small, incremental changes at a time. :grin:

By the end of the discussion in #448, it seemed we were leaning towards the Dealer returning an ERROR message with the `wamp.error.cancelled` URI. This is the current behavior for Crossbar. The other alternative was returning a more-specific URI such as `wamp.error.callee_left`.

```mermaid
sequenceDiagram
    participant Caller
    participant Dealer
    participant Callee

    Caller->>Dealer: CALL
    Dealer->>Callee: INVOCATION
    Callee->>Dealer: GOODBYE
    Dealer->>Caller: CALL ERROR (cancelled)
```

If the Caller never sent a CANCEL message, then it can deduce that the `wamp.error.cancelled` was caused by the Callee leaving.

If the Caller did send a CANCEL message, it cannot know if the cancellation was due to normal cancellation, or due to the Callee leaving (unless ERROR payload args are used to convey more information). In either case, the desired effect to cancel the call is the same from the point of view of the Caller.

I propose that the following text be added to the _Remote Procedures Calls_ section of the basic profile:

> ## 6.3 Abnormal Situations
>
> ### 6.3.1 Callee Leaving Before Call Completion
>
> After sending an `INVOCATION` message, if a *Dealer* detects that the *Callee* has terminated its session without sending a `YIELD` or `ERROR` response, then the *Dealer* MUST return an `ERROR` message back to the *Caller* with a `wamp.error.cancelled` URI. The *Dealer* MAY provide additional information via the `ERROR` payload arguments to clarify that the cancellation is due to the *Callee* leaving before the call could be completed.
>
> (insert above sequence diagram)

Another subsection should be added for Caller leaving, but I prefer to create a separate issue to discuss that.

Furthermore, the `wamp.error.cancelled` ERROR URI would be moved from the Advanced profile to the Basic profile.

ecorm:
Resolves #476 and #477

Pinging @oberstet @KSDaemon @Jopie64 for review

### DIFF
--- a/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
+++ b/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
@@ -337,15 +337,15 @@ The *Dealer*'s behavior for when a *Callee* leaves or disconnects during a progr
         |                  |                 |
         | RESULT (progress)|                 |
         | <----------------|                 |
-        |                  |     GOODBYE     |
-        |                  |<--------------- |
-        |                  |                 |
-        |      ERROR       |                 |
-        | <--------------- |                 |
-        |                  |                 |
-     ,--+---.           ,--+---.          ,--+---.
-     |Caller|           |Dealer|          |Callee|
-     `------'           `------'          `------'
+        |                  |              ,--+---.
+        |                  |              |Callee|
+        |                  |              `------'
+        |      ERROR       |               (gone)
+        | <--------------- |
+        |                  |
+     ,--+---.           ,--+---.
+     |Caller|           |Dealer|
+     `------'           `------'
 
 
 **Ignoring Requests for Progressive Call Results**
--- a/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
+++ b/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
@@ -312,8 +312,11 @@ The *Dealer*'s behavior for when a *Caller* leaves or disconnects during a progr
      `------'              |    INTERRUPT    |
       (gone)               | ---------------->
                            |                 |
+                        ,--+---.          ,--+---.
+                        |Dealer|          |Callee|
+                        `------'          `------'
 
-Such cancellation of Progressive Call Results addresses a potential security vulnerability: In cases where progressive results are used to stream data to *Callers*, and network connectivity is unreliable, *Callers* may often get disconnected in the middle of receiving such progressive results. Without the mandated cancellation behavior, recurring connect-call-disconnect cycles by a *Caller* would result in a rapidly growing backlog of unprocessed streaming results, overloading the router and further degrading network connectivity.
+Such cancellation when the caller leaves addresses a potential security vulnerability: In cases where progressive results are used to stream data to *Callers*, and network connectivity is unreliable, *Callers* may often get disconnected in the middle of receiving such progressive results. Without the mandated cancellation behavior, recurring connect-call-disconnect cycles by a *Caller* would result in a rapidly growing backlog of unprocessed streaming results, overloading the router and further degrading network connectivity.
 
 
 **Callee Leaving**
--- a/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
+++ b/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
@@ -289,20 +289,9 @@ The progressive `YIELD` and progressive `RESULT` may also be empty, e.g. when th
 
 Even if a *Caller* has indicated its expectation to receive progressive results by setting `CALL.Options.receive_progress|bool := true`, a *Callee* is **not required** to produce progressive results. `CALL.Options.receive_progress` and `INVOCATION.Details.receive_progress` are simply indications that the *Caller* is prepared to process progressive results, should there be any produced. In other words, *Callees* are free to ignore such `receive_progress` hints at any time.
 
-**Progressive Call Result Cancellation**
+**Caller Leaving**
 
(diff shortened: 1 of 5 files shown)


SunFulong:
What should Dealer do if:
1. Dealer response to Caller: Error (Timeout)
2. Callee Response to Dealer: Yield
3. Dealer Response to Callee: ?

ecorm:
> What should Dealer do if...

The [Dealer Initiated Timeouts](https://wamp-proto.org/wamp_latest_ietf.html#section-11.3-5) subsection specifies that:

> the message flow is identical to Call Canceling, except that there is no CANCEL message that originates from the Caller. The cancellation mode is implicitly `killnowait` if the Callee supports call cancellation, otherwise the cancellation mode is `skip`.

The description for `skip` and `killnowait` modes in the [Call Canceling](https://wamp-proto.org/wamp_latest_ietf.html#name-call-canceling) section specify that an ERROR is returned immediately and that YIELD or ERROR results sent by the Callee are discarded by the Dealer.

I suppose that a sentence or two could be added to the Dealer Initialiated Timeouts subsection to clarity this without having to piece together the desired behavior from 2-3 places.

But I'm afraid your question is not related to this PR which deals with the Caller or Callee leaving during a CALL. I want to keep the scope of this PR small to avoid endlessly long discussions. Please create a separate issue if you feel the spec is lacking with regards to your question.

SunFulong:
> But I'm afraid your question is not related to this PR which deals with the Caller or Callee leaving during a CALL. I want to keep the scope of this PR small to avoid endlessly long discussions.

Right, I will do it.

rfc/text/advanced/ap_09_rpc_progressive_call_results.md
```diff
@@ -321,18 +309,41 @@ The message flow for progressive results cancellation involves:
...
-      (gone)               | <----------------
-                           |                 |
-                           |    INTERRUPT    |
-                           | ---------------->
+     `------'              |    INTERRUPT    |
+      (gone)               | ---------------->
                            |                 |
-                        ,--+---.          ,--+---.
-                        |Dealer|          |Callee|
-                        `------'          `------'
```
KSDaemon, line 332: Seems this was accidentally deleted

rfc/text/advanced/ap_09_rpc_progressive_call_results.md
```diff
@@ -289,28 +289,16 @@ The progressive `YIELD` and progressive `RESULT` may also be empty, e.g. when th
...
-```
-Options:
-```
-INTERRUPT.Options.mode|string == "killnowait"
-```
-
-Cancellation of Progressive Call Results addresses a potential security vulnerability: In cases where progressive results are used to stream data to *Callers*, and network connectivity is unreliable, *Callers* may often get disconnected in the middle of receiving such progressive results. Without the mandated cancellation behavior, recurring connect-call-disconnect cycles by a *Caller* would result in a rapidly growing backlog of unprocessed streaming results, overloading the router and further degrading network connectivity.
-
-The message flow for progressive results cancellation involves:
+The *Dealer*'s behavior for when a *Caller* leaves or disconnects during a progressive results call shall be the same as in a basic, non-progressive call. That is, the *Dealer* sends an INTERRUPT to the *Callee* with `mode="killnowait"`. See [Caller Leaving During RPC Invocation] (#rpc-caller-leaving) under the Basic Profile.
```
KSDaemon, line 294: Maybe its worth noting that if Callee had time to send a YIELD message before it received INTERRUPT from the Dealer, the DEALER has nothing to do with the YIELD message and just omits it.

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
```diff
@@ -174,9 +157,34 @@ The message flow for cancelling calls with progressive call invocations involves
                          |Dealer|                 |Callee|
                          `------'                 `------'
 
+Progressive call cancellation, like in progressive call invocations, addresses a potential security vulnerability:
```
KSDaemon, line 160:
```suggestion
Progressive call invocations cancellation, like in progressive call results, addresses a potential security vulnerability:
```
I suppose you mean this?

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
```diff
@@ -174,9 +157,34 @@ The message flow for cancelling calls with progressive call invocations involves
...
+        |                  |                       |
+        |                  |                       |
+        |                  |        GOODBYE        |
+        |                  |<--------------------- |
+        |      ERROR       |                       |
+        | <--------------- |                       |
+        |                  |                       |
+     ,--+---.           ,--+---.                ,--+---.
+     |Caller|           |Dealer|                |Callee|
+     `------'           `------'                `------'
```
KSDaemon, line 187: I think this should looks like other examples: marking the Callee as `gone`

KSDaemon (commented):
Looks good! Just a few minor notes

rfc/text/advanced/ap_09_rpc_progressive_call_results.md
KSDaemon: sorry, missed that.

KSDaemon (approved):
Do not see any reason against merging this! Awesome! 👍🏻

ecorm:
Sorry, but, bump?

KSDaemon:
Oh sorry! Thnx for pinging! As there is no words against — merging! Thnx @ecorm !

PR #483, opened 2023-05-17
Use wamp.close.killed when session forcefully killed via meta API

### LINKED ISSUE #462: Consider renaming wamp.close.normal
If it's not too late, I would like to discuss the rationale for the naming behind the `wamp.close.normal` error URI, used as default for the [wamp.session.kill](https://wamp-proto.org/wamp_latest_ietf.html#name-wampsessionkill) meta procedure.

From the point of view of the client, having one's session forcibly killed is certainly not "normal". I would suggest something like `wamp.close.killed` instead.

I apologize for missing this back then I reviewed PR #315.

ecorm:
Fixes #462


oberstet (approved):

KSDaemon (approved):
👍🏻 Great! Thnx!

PR #491, opened 2023-05-30
Contuations on completed progressive calls

### LINKED ISSUE #482: Error URI for expired progressive call race condition
This is a continuation of the long discussion in #448, but will focus purely on the race conditions where a progressive CALL is made without the Caller knowing that the RPC is completed.

I can think of two scenarios where this can happen:
1. The callee left before the RPC was completed
2. The callee already returned an ERROR or a final RESULT

Here's a sequence diagram of the first scenario (where Network is used to simulate delay):

```mermaid
sequenceDiagram
    participant Caller
    participant Network
    participant Dealer
    participant Callee

    Caller->>Network: CALL (progress=true)
    Network->>Dealer: CALL (progress=true)
    Dealer->>Callee: INVOCATION
    Caller->>Network: CALL (progress=true)
    Callee->>Dealer: GOODBYE
    Dealer->>Network: ERROR (canceled) [1]
    Dealer->>Dealer: Delete call request record
    Network->>Caller: ERROR (canceled)
    Network->>Dealer: CALL (progress=true)
    Note right of Dealer: Call request ID not found
    Dealer->>Network: ERROR (which URI?)
    Network->>Caller: ERROR (which URI?)
```
[1] This behavior [is now mandated in the spec](https://wamp-proto.org/wamp_latest_ietf.html#name-callee-leaving-during-an-rp) as of PR #478 .

And here's a sequence diagram of the second scenario:

```mermaid
sequenceDiagram
    participant Caller
    participant Network
    participant Dealer
    participant Callee

    Caller->>Network: CALL (progress=true)
    Network->>Dealer: CALL (progress=true)
    Dealer->>Callee: INVOCATION
    Caller->>Network: CALL (progress=true)
    Callee->>Dealer: YIELD (progress=false)
    Dealer->>Network: RESULT (progress=false)
    Dealer->>Dealer: Delete call request record
    Network->>Caller: RESULT (progress=false)
    Network->>Dealer: CALL (progress=true)
    Note right of Dealer: Call request ID not found
    Dealer->>Network: ERROR (which URI?)
    Network->>Caller: ERROR (which URI?)
```

Both sequences are very similar. In both cases, the router receives a second progressive CALL with a request ID for which it cannot find a matching RPC in progress. It would be inappropriate to treat this as a protocol violation, because this is due to a race condition that is outside the control of the Caller.

**So I believe the question remains: what should be the error URI for these cases?**

There is `wamp.error.no_such_procedure`, but that would be a "lie". In the first scenario, the procedure could still be registered to another Callee via the shared registrations feature. In the second scenario, the procedure could still be registered to the same Callee.

I think that leaves `wamp.error.canceled` as the only viable candidate if we limit ourselves to existing error URIs. I think `wamp.error.canceled` would confusing in the second scenario, because the call was actually completed, not cancelled.

I therefore suggest: `wamp.error.no_such_call`. It would fit the theme for similarly-named URIs (e.g. `no_such_procedure`, `no_such_registration`, etc).

Another alternative would be not returning an ERROR at all, and ignore the second bogus CALL with no matching request ID. But I'm worried this could be lead to deadlock situations where a Caller is waiting forever for a response. I think it's better to return *something* and not leave the Caller waiting.

ecorm:
Fixes #482

Pinging @Jopie64 , @KSDaemon , @oberstet for review.

### DIFF
--- a/rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
+++ b/rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
@@ -189,9 +189,9 @@ The *Dealer*'s behavior for when a *Callee* leaves or disconnects during a call 
 
 **Continuations on Completed Calls**
 
-A call with progressive invocations is considered completed by the *Dealer* when the latter receives a correlated YIELD or ERROR message from the *Callee*, or when the *Callee* leaves the realm.
+A call with progressive invocations is considered completed by the *Dealer* when the latter receives a correlated final YIELD or ERROR message from the *Callee*, or when the *Callee* leaves the realm.
 
-Due to network delays, a *Caller* may be unaware that the call is completed by time it sends a progressive CALL continuation.
+Due to network delays, the *Caller* may be unaware that the call is completed by time it sends another progressive CALL continuation.
 
 When a *Dealer* receives a CALL under the following conditions:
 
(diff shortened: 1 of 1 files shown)


Jopie64 (commented):
Looks good I think...
Just realized the same race could happen at the callee end: it receives an innovation while it already sent an error/yield... I guess that should be mentioned as well?

ecorm:
> Looks good I think... Just realized the same race could happen at the callee end: it receives an innovation while it already sent an error/yield... I guess that should be mentioned as well?

Good catch; I think we should include that as well. So the Callee should just ignore it, then?

Jopie64:
> Good catch; I think we should include that as well. So the Callee should just ignore it, then?

Yea, similar as with dealer.

oberstet (approved):

KSDaemon:
I'll try to look in the evening.

ecorm:
Added text for Callee progressive invocation race

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
```diff
@@ -234,9 +234,53 @@ The sequence diagram below illustrates this sitation, where the Network actor mo
...
+
+Due to network delays, the *Dealer* may be unaware that the call is completed by time it sends another progressive INVOCATION.
+
+When a *Callee* receives an INVOCATION under the following conditions:
+
+- the *Callee* supports Progressive Call Invocations,
+- the INVOCATION request ID does not correspond to a new call request, and,
+- the INVOCATION request ID does not match any RPC invocation in progress,
+
+then it MUST ignore and discard that INVOCATION message without any further correlated response to the *Dealer*. The rationale for this is that the *Caller* will eventually receive a correlated RESULT or ERROR message from the previous call completion and will be able to handle the call completion accordingly.
```
Jopie64, line 250: *Caller* should be *dealer* and RESULT should be YIELD here right?

KSDaemon (approved):
Great Addition! 👍🏻  Thnx @ecorm

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
KSDaemon: Caller should be dealer — right! And fixed! 
But RESULT should be RESULT as it is about Caller side in the end and not about Dealer side. So I think it is correct.

Jopie64: But if you changed caller into dealer, but not RESULT into YIELD, then now it reads
> ... the *dealer* will eventually receive a correlated RESULT ...

And that's not correct right?

I think in this part the rationale is supposed to be about the dealer who doesn't need an extra ERROR for the raced INVOCATION since it will eventually always receive the ERROR/YIELD that has been delayed by the network...

KSDaemon: Sorry @Jopie64, it seems that I said it incorrectly. I reread it again. The final merged version:
```
The rationale for this is that the *Caller* will eventually receive a correlated RESULT or ERROR message from the previous call completion and will be able to handle the call completion accordingly.
```

Looks to be clear and correct. Or do you think that is not?

Jopie64: Not sure... Also not sure how I could see the final version on my mobile 😅

There are two places with this rationale. One is about race between caller-dealer, and one between dealer-callee.

The one you quoted would be correct for the one between caller-dealer, but I thought this part was about the race between dealer-callee...

KSDaemon: Yeah. But in any case, the call ends on the Caller's side with the RESULT or ERROR. And that is stated. If I correctly understand you, you want to extend the description for the dealer-callee race with smth like that:

> The rationale for this is that the Dealer will eventually receive a correlated YIELD or ERROR message from the Callee and then send a correlated RESULT or ERROR message to the Caller so it will be able to handle the call completion accordingly.

Right?

ecorm: > Yeah. But in any case, the call ends on the Caller's side with the RESULT or ERROR. And that is stated. If I correctly understand you, you want to extend the description for the dealer-callee race with smth like that:
> 
> > The rationale for this is that the Dealer will eventually receive a correlated YIELD or ERROR message from the Callee and then send a correlated RESULT or ERROR message to the Caller so it will be able to handle the call completion accordingly.
> 
> Right?

I prefer this version of the paragraph. Shall I submit another PR with this change?

ecorm:
Submitted #492, which contains the changes requested by @Jopie64 and drafted by @KSDaemon .

Issue #448, opened 2023-02-22
Calling Expired Progressive Calls


ecorm:
What happens when a CALL message uses a request ID for a progressive call that no longer exists (due to the callee unregistering or leaving) and the caller has not yet received the `wamp.error.cancelled` error? I don't think aborting with `wamp.error.protocol_violation` would be appropriate here, because the race condition is outside the caller's control.

I propose that in such scenarios, the dealer responds with an error having a `wamp.error.no_such_call` URI. `wamp.error.no_such_procedure` would not be appropriate here because the procedure may still exist - it's merely the progressive call that has expired.

Note that when progressive calls are supported, a router can no longer simplistically reject CALL request IDs that are not sequential. The router must allow the caller to use an old request ID that may reference a recently expired progressive call. It would be too costly for a router to remember all previous progressive call request IDs.

There is therefore a scenario where a CALL uses an old request ID for a request that is still active, but is not a progressive call. In those instances, I propose that they be treated as protocol violations.

(This issue is for the continuation of a discussion I started in PR #445)

KSDaemon:
Definitely, a situation when a router receives a call with a request ID of an ongoing non-progressive call should be treated as `protocol_violation`. That can not happen in a legitime flow.

The case when a router receives a progressive call with a request ID that doesn't exists in the router runtime can be treated as a starting new progressive call. The only way to distinguish whether is it an obsolete progressive call request (that was canceled, callee-leaved, etc) or a brand-new one — is to analyze the request id: is it the next one or previously used). 

I've looked into Nexus Router implementation — it doesn't keep track of the last used request ID, only checks for existence. 

I dug through the spec and can't find a description of what should happen if one of the peers (mostly router) receives a non-incremental request ID like previously used or +N (N>1) to current. Am I missing it or we indeed didn't describe it? 

So the questions are:

* What the peer should do when it receives a non-incremental request ID (in session scope)?
*  How do current router implementations work? Do they abort the connection? Nexus — doesn't care. (I don't say it is good or bad, just fact)
* Depending on this we can think of how to deal with the original question.

ecorm:
> KSDaemon: The case when a router receives a progressive call with a request ID that doesn't exists in the router runtime can be treated as a starting new progressive call. The only way to distinguish whether is it an obsolete progressive call request (that was canceled, callee-leaved, etc) or a brand-new one — is to analyze the request id: is it the next one or previously used).

That's exactly how I do it in my work-in-progress router implementation.

[...]

PR #495, opened 2023-06-16
Tolerate non-sequential CALL request IDs

### LINKED ISSUE #494: Tolerating old CALL request IDs when progressive invocations are supported
This is a continuation of the long discussion in #448, but will focus purely on the aspect of CALL request IDs for invocations that no longer exist.

Section [2.3.3 Protocol errors](https://wamp-proto.org/wamp_latest_ietf.html#name-protocol-errors) currently prohibits CALL request IDs that are not sequential:

> Receiving message with non-[sequential](https://wamp-proto.org/wamp_latest_ietf.html#session_scope_id) [session scope](https://wamp-proto.org/wamp_latest_ietf.html#session_scope_ids) request ID, such as SUBSCRIBE, UNSUBSCRIBE, PUBLISH, REGISTER, UNREGISTER, CALL and YIELD.

This needs to be fixed so that CALL request IDs may refer to progressive invocation calls and need not be strictly sequential. It would be more accurate to state that **new** CALL request IDs must be sequential.

I'm going to introduce a term here to simplify discussion. The CALL request ID _watermark_ is the value of highest valid CALL request ID that has been received by the Dealer.

Another term I'll introduce is a _call continuation_, which is a subsequent CALL to a progressive invocation call request (which may or may not be final). A call continuation has a request ID that is less than or equal to the call request ID watermark.

When a Dealer receives a CALL request ID that is less than or equal to the watermark, it may want to verify that the CALL is an actual call continuation. The problem is that the progressive invocation call may be already complete by time it receives the continuation (we covered this race condition in #482). So if the Dealer wants to perform this verification, it must remember previous request IDs associated with progressive invocation calls.

I think imposing the cost of remembering the request IDs of previous progressive calls is too much to ask. Sure, the Dealer could just remember them for a limited time, but how much time is appropriate?

I therefore propose that Section 2.3.3 (Protocol errors) be amended as follows:

> Receiving message with non-[sequential](https://wamp-proto.org/wamp_latest_ietf.html#session_scope_id) [session scope](https://wamp-proto.org/wamp_latest_ietf.html#session_scope_ids) request ID, such as SUBSCRIBE, UNSUBSCRIBE, PUBLISH, REGISTER, UNREGISTER, ~CALL and YIELD~.
>
> In the case of CALL messages, only those which initiate a new RPC shall be required to have a sequential session scope request ID. For example, when the _Progressive Invocation Calls_ feature is supported, a CALL request ID may less than or equal to the highest valid request ID sent by the *Client*. A *Dealer* MAY choose to remember request IDs of previously completed progressive invocations calls for verification purposes, but this is not required.

The requirement that YIELD messages have sequential request IDs should be removed. The Callee is not generating a request ID for a YIELD message, but is instead correlating with the INVOCATION sent by the Dealer.

ecorm:
Fixes #494
Pinging @Jopie64 , @KSDaemon , @oberstet for review.


Jopie64 (approved):
Looks good to me! 👍
Do we have consensus over this? Also from @oberstet ?

ecorm:
> Do we have consensus over this? Also from @oberstet ?

Nexus is the only released router which currently supports progressive call invocations (to my knowledge), so I would be good if @KSDaemon also takes a look at it before merging.

KSDaemon (approved):
It looks clear to me! 👍🏻

oberstet:
> Therefore, when progressive call invocations are enabled, the requirement for CALL request IDs to be sequential session scope is waived (see [Protocol Violations]

yes, agreed. actually, reusing the request ID of the original CALL in subsequent progressive CALL messages is the mechanism to recognize the subsequent messages to be progress for the original CALL

btw, do we want to use language like: the whole series of messages (in both directions) is called one call _action_?

> Because a Router is not expected to remember every completed request, a CALL may even use the request ID of a previously **completed** non-CALL request, and it would not be considered a protocol violation.

this aspect is unrelated to the "tolerate same request ID wrt progress" ... it is also about the strictness of sequentiality checks

if we would allow that, that would mean: *request IDs are no longer necessarily *monotonically increasing without gaps (sequential) for WAMP actions (and identical within one action).*

do we really want to go there?

ecorm:
> btw, do we want to use language like: the whole series of messages (in both directions) is called one call action?

I think "transaction" or "transfer" would be a better term if were to introduce new language.

> this aspect is unrelated to the "tolerate same request ID wrt progress" ... it is also about the strictness of sequentiality checks
>
> if we would allow that, that would mean:
>
> request IDs are no longer necessarily monotonically increasing without gaps (sequential) for WAMP actions (and identical within one action).
>
> do we really want to go there?

It is only for CALLs that are not new transactions that the requirement for monoticity is waived, and only when progressive call invocations are enabled.

The only way for a router to be strict in terms of CALL request IDs, is to implement a "grace period" after a progressive call is completed where a caller can reuse old CALL request IDs. The router would have "remember" the request IDs of progressive invocation calls for their respective grace periods. We could mention this approach in the spec, and write that routers SHOULD implement this (but not MUST).

ecorm:
I rewrote the text so that strict verification of CALL request IDs can still be done if a router chooses to. I will also add some pseudocode for the request ID verification algorithm.

ecorm:
@Jopie64 , @KSDaemon , @oberstet Can you check out the changes I made to this PR which allows strict checking of CALL request IDs?

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
```diff
@@ -282,6 +282,25 @@ The sequence diagram below illustrates this sitation, where the Network actor mo
      `------'     `------'        `-------'       `------'
 
 
+**Non-sequential CALL Request IDs**
+
+When sending a CALL that continues a progressive call invocation, its request ID is the same as the CALL that initiated the progressive call invocation. Therefore, when progressive call invocations are enabled, the requirement for CALL request IDs to be sequential session scope is waived (see [Protocol Violations](#protocol_errors)). Only those CALL messages which initiate a new RPC shall be required to have a sequential session scope request ID. A Router MAY choose to remember request IDs of previously completed progressive invocations calls for verification purposes, but this is not required.
```
Jopie64: > A Router MAY choose to remember request IDs of previously completed progressive invocations calls for verification purposes, but this is not required.

Router -> Dealer

Also I think it's worth noting that if a dealer wants to verify whether a non-sequential id is from a previously completed progressive call, care must be taken to find the right balance in the time these id's are remembered. Too short and you can have data races with slow networks and wrongly flag invalid call id's. Too long and you might end up with a memory overflow especially when lots of progressive calls are made.

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
```diff
@@ -282,6 +282,25 @@ The sequence diagram below illustrates this sitation, where the Network actor mo
...
+
+For example, this sequence of CALL messages received by a Router is considered valid in terms of protocol violation detection:
+
+1. `[CALL, 1, {"progress":true}, "foo"]`
+2. `[CALL, 2, {}, "bar"]`
+3. `[CALL, 1, {}, "foo"]`
+
+Because a Router is not expected to remember every completed request, a CALL may even use the request ID of a previously **completed** non-CALL request, and it would not be considered a protocol violation. In this case, the Router shall simply discard the CALL message, as already discussed in the previous section.
+
+However, if the Router receives a CALL for a _new_ RPC that skips the next expected request ID, then that shall be considered a protocol violation:
```
Jopie64: Maybe for clarity:
... then that shall _always_ be considered a protocol violation, even in cases where progressive calls are made.

Jopie64 (commented):
Looks good! Just a few remarks

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
Jopie64: Also router -> dealer here 😁

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
Jopie64: > When sending a CALL that continues a progressive call invocation

"invocation" seems redundant here. I think you can leave it out.

ecorm: It appears you reviewed an older version of the text, but I applied your suggestions to the current version in this PR.

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
ecorm: It appears you reviewed an older version of the text, but I applied your suggestions to the current version in this PR.

rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
Jopie64: Yea it looks like I did. I'm still getting used to reviewing using GitHub since I'm used to a different system... Not sure how to review multiple commits of the same pr at the same time yet, but managed to view your latest changes. And to me it looks good now 😊

ecorm: The next time, you can go to the Files Changes tab where it shows the changes for all commits at once. You can then click on a + or - to comment and start a review.

oberstet:
@ecorm sorry for my sluggish response time! this looks great, I like the detailed example showcasing the problem resulting from the old, non-refined sequence number rules. and "RPC transactions" etc. awesome, let's merge!

ecorm:
@oberstet No worries. It's summer and it's understandable that everyone is probably super busy. This completes the series of sub-issues that was raised by issue #448.

PR #498, opened 2023-06-19
Replaced all NBSP with spaces

KSDaemon:
Replaced all `NBSP` (that results in (U+00A0) after rendering) with just spaces. So now the spec should look clean

### DIFF
--- a/_work/predefined-uris.md
+++ b/_work/predefined-uris.md
@@ -2,11 +2,11 @@
 
 # Predefined URIs
 
-WAMP predefines the following URIs in the *advanced profile*. For URIs, used in *basic profile*, please, see appendix in basic profile specification.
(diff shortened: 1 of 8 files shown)


oberstet (approved):

KSDaemon:
I've just noticed one of them 'cause JetBrains IDEs are showing them specifically. It might be some "clever" editors like JB IDEs. Maybe at that point that got into the spec and I missed that. Because I turned off that cleverness the first time I mentioned it :)

KSDaemon:
@SunFulong :) I'm glad!) 

> BTW, could we remove each code block leading spaces?

That can be tricky. Afaik, if we do that - it will break the RFC compilation and the result will be broken. Need to check that.

PR #501, opened 2023-07-05
REGISTER.Options.forward_timeout option

### LINKED ISSUE #500: Add `REGISTER.Options.forward_timeout` option
Call timeouts can be executed in three places: the caller, the dealer, or the callee.

In the case of caller-initiated timeouts, there are no extra options added to CALL, and the caller handles the timeout logic itself, sending a CANCEL message if it wants to cancel the call when a timeout occurs.

In the case of dealer or callee-initiated timeouts, a `CALL.Options.timeout` option is added. The spec currently has this to say about which peer handles the timeout:

> The decision by *Dealers* to delegate timeouts to *Callees* supporting this feature MAY be determined by router configuration.

Instead of determining by router configuration (or in addition to the configuration), I propose that an option be added to REGISTER so that a callee may specify its desire to receive the `timeout` option and be responsible for handling timeouts. This would allow the Dealer's timeout forwarding decision to be more granular, that is, on a per-procedure basis (instead of all-or-nothing).

I'd be willing to write the PR for this proposed enhancement, but I first want to make sure the idea is supported.

ecorm:
Fixes #500.
Pinging @Jopie64 who commented in the related issue. Anyone else is welcome to comment/review of course.

### DIFF
--- a/rfc/text/advanced/ap_11_rpc_call_timeout.md
+++ b/rfc/text/advanced/ap_11_rpc_call_timeout.md
@@ -19,7 +19,9 @@ The error message that is returned to the *Caller* MUST use `wamp.error.timeout`
 
 If the *Callee* supports Call Timeouts, the *Dealer* MAY propagate the `CALL.Options.timeout|integer` option via the `INVOCATION` message and allow the *Callee* to handle the timeout logic. If the operation times out, the *Callee* MUST return an `ERROR` message with `wamp.error.timeout` as the reason URI.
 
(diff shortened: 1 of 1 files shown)


ecorm:
Feel free to bikeshed the `REGISTER.Options.forward_timeout` option name.

KSDaemon (approved):
👍🏻 I like it

oberstet (approved):

ecorm:
Sorry, but... bump?

KSDaemon:
Sure! Merging!

PR #506, opened 2023-10-24
Abort with payload

### LINKED ISSUE #499: Add Payload to Abort
`ABORT` currently only has a message dictionary that is part of the head of the message. If an `ABORT` needs to send rich content or private content to the client receiving the `ABORT`. The message dictionary need to be used at the moment. The problem is that some advanced and custom features like

- opaque payload
- E2E encryption
- zero-copy routing

will be impossible, leading to an unsecure or slow data transfer. Thats why it would be great to have a Payload added to `ABORT`.

konsultaner:
fixes #499

### DIFF
--- a/rfc/text/basic/bp_03_messages.md
+++ b/rfc/text/basic/bp_03_messages.md
@@ -75,6 +75,10 @@ Sent by a Peer*to abort the opening of a WAMP session. No response is expected.
 {align="left"}
         [ABORT, Details|dict, Reason|uri]
 
+        [ABORT, Details|dict, Reason|uri, Arguments|list]
+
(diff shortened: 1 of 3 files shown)


KSDaemon (approved):
👍🏻  Looks good at first glance.

oberstet:
The PR looks good, it reflects the result of our discussion! 

> I have reread parts of the discussion .. again stumbled across the same "asymmetry": even though https://github.com/wamp-proto/wamp-proto/issues/499#issuecomment-1605006410 is still true IMO, the result of the discussion was positive to add args/kwargs to ABORT - and only to ABORT - https://github.com/wamp-proto/wamp-proto/issues/499#issuecomment-1621820938

PR #508, opened 2023-11-07
Fix raw socket maximum length limit

### LINKED ISSUE #507: Incorrect maximum raw socket limit
The spec specifies `2**24 octets` as the upper limit for the handshake length field, but the header only supports a length up to 0x00FFFFFF = 2^24-1 octets.

ecorm:
Fixes #507

### DIFF
--- a/rfc/text/advanced/ap_38_transport_rawsocket.md
+++ b/rfc/text/advanced/ap_38_transport_rawsocket.md
@@ -63,9 +63,10 @@ The possible values for `LENGTH` are:
          0: 2**9 octets
          1: 2**10 octets
         ...
+        14: 2**23 octets
         15: 2**24 - 1 octets
 
-This means a *Client* can choose the maximum message length between **512** and **16M-1** octets.
+This means a *Client* can choose the maximum message length between **512** and **16M-1** octets. In the case where the `LENGTH` field is 15, 1 must be subtracted from the 16M octet length because otherwise it would not fit in the 24 bits allotted to frame length in the message frame header (see the Framing subsection below).
 
 The `SERIALIZER` value is used by the *Client* to request a specific serializer to be used. When the handshake completes successfully, the *Client* and *Router* will use the serializer requested by the *Client*.
 
--- a/rfc/text/advanced/ap_38_transport_rawsocket.md
+++ b/rfc/text/advanced/ap_38_transport_rawsocket.md
@@ -242,7 +242,7 @@ The four bits `RRRR` are reserved for future use and MUST be all zeros for now.
         31                                    0
         RRRR 1TTT 0000 0000 0000 0000 0000 0000
 
-When the payload length is 16M-1 octets, the prefix would be:
+When the payload length is 16M-1 octets, for example, the prefix would be:
 
 {align="left"}
         MSB                                 LSB
--- a/rfc/text/advanced/ap_38_transport_rawsocket.md
+++ b/rfc/text/advanced/ap_38_transport_rawsocket.md
@@ -63,9 +63,9 @@ The possible values for `LENGTH` are:
          0: 2**9 octets
          1: 2**10 octets
         ...
-        15: 2**24 octets
+        15: 2**24 - 1 octets
 
-This means a *Client* can choose the maximum message length between **512** and **16M** octets.
+This means a *Client* can choose the maximum message length between **512** and **16M-1** octets.
 
 The `SERIALIZER` value is used by the *Client* to request a specific serializer to be used. When the handshake completes successfully, the *Client* and *Router* will use the serializer requested by the *Client*.
 
@@ -117,7 +117,10 @@ for i in range(256):
     ser_id = i & 0x0f
     if ser_id != 0:
         ser = SERMAP.get(ser_id, 'currently undefined')
-        maxlen = 2 ** ((i >> 4) + 9)
+        lenbits = i >> 4
+        maxlen = 2 ** (lenbits + 9)
+        if lenbits == 15
+            maxlen -= 1
         print("{:02x} => serializer: {}, maxlen: {}".format(i, ser, maxlen))
     else:
         print("fail the connection: illegal serializer value")
@@ -192,7 +195,10 @@ for i in range(256):
         # verify the serializer is the one we requested!
         # if not, fail the connection!
         ser = SERMAP.get(ser_id, 'currently undefined')
-        maxlen = 2 ** ((i >> 4) + 9)
+        lenbits = i >> 4
+        maxlen = 2 ** (lenbits + 9)
+        if lenbits == 15
+            maxlen -= 1
         print("{:02x} => serializer: {}, maxlen: {}".format(i, ser, maxlen))
     else:
         err = i >> 4
--- a/rfc/text/advanced/ap_38_transport_rawsocket.md
+++ b/rfc/text/advanced/ap_38_transport_rawsocket.md
@@ -118,10 +118,7 @@ for i in range(256):
     ser_id = i & 0x0f
     if ser_id != 0:
         ser = SERMAP.get(ser_id, 'currently undefined')
-        lenbits = i >> 4
-        maxlen = 2 ** (lenbits + 9)
-        if lenbits == 15
-            maxlen -= 1
+        maxlen = 2 ** ((i >> 4) + 9)
         print("{:02x} => serializer: {}, maxlen: {}".format(i, ser, maxlen))
     else:
         print("fail the connection: illegal serializer value")
@@ -196,10 +193,7 @@ for i in range(256):
         # verify the serializer is the one we requested!
         # if not, fail the connection!
         ser = SERMAP.get(ser_id, 'currently undefined')
-        lenbits = i >> 4
-        maxlen = 2 ** (lenbits + 9)
-        if lenbits == 15
-            maxlen -= 1
+        maxlen = 2 ** ((i >> 4) + 9)
         print("{:02x} => serializer: {}, maxlen: {}".format(i, ser, maxlen))
     else:
         err = i >> 4
--- a/rfc/text/advanced/ap_38_transport_rawsocket.md
+++ b/rfc/text/advanced/ap_38_transport_rawsocket.md
@@ -64,9 +64,9 @@ The possible values for `LENGTH` are:
          1: 2**10 octets
         ...
         14: 2**23 octets
-        15: 2**24 - 1 octets
+        15: 2**24 octets
 
-This means a *Client* can choose the maximum message length between **512** and **16M-1** octets. In the case where the `LENGTH` field is 15, 1 must be subtracted from the 16M octet length because otherwise it would not fit in the 24 bits allotted to frame length in the message frame header (see the Framing subsection below).
+This means a *Client* can choose the maximum message length between **512** and **16M** octets.
 
 The `SERIALIZER` value is used by the *Client* to request a specific serializer to be used. When the handshake completes successfully, the *Client* and *Router* will use the serializer requested by the *Client*.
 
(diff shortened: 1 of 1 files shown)


KSDaemon (approved):
👍🏻 Thnx @ecorm for catching this!

ecorm:
I just occurred to me that it would be more consistent if `-1` was applied to all the other cases as well, but that would probably break some implementations. Also, buffer sizes tend to be at exact powers of two.

oberstet:
first of all, thanks a lot for having such a detailed look at the spec and also (in this case) the autobahn implementation!!

having said that, I do think the current spec and implementation is "correct" (makes sense, is consistent).

IMO, you might mix up "last index" and "buffer size".

with `LENGTH == 15`, the maximum message (payload) length is 2**24 octets (without -1), BUT within a buffer of such a size, the last _index_ of an octet as an offset from the first octet (the buffer start) is 2**24 - 1.

so actually I don't see the issue in https://github.com/wamp-proto/wamp-proto/issues/507 .. could you please explain/expand? maybe I'm just too dumb to get it the issue/problem .. this happens, no question;)

ecorm:
@oberstet

> BUT within a buffer of such a size, the last index of an octet as an offset from the first octet (the buffer start) is 2**24 - 1. it's the old "zero-indexed" vs "numbered" thing. C is zero indexed, math uses numbering (starting with 1).

I'm afraid this has nothing to do with indexing. I'm talking about the **length** field in the rawsock frame header, where 0 means "there is no payload" and 3 means "the payload is three bytes long". I have never encountered a binary protocol where a length field is offset by one and does not represent the actual length in bytes.

There are only 24 length bits in the raw socket frame headers. So if you specify a frame length of `2^24 = 16777216 = 0x1000000 = 0b1'0000'0000'0000'0000'0000'0000`, it won't fit in the 24 bits allotted for the length.

```
    MSB                                 LSB
    31                                    0
    RRRR RTTT LLLL LLLL LLLL LLLL LLLL LLLL
    0000 0001 0000 0000 0000 0000 0000 0000
            ^
          Oops!
```

If we try with a length of `2^24 - 1 = 16777215 = 0xFFFFFF = 0b1111'1111'1111'1111'1111'1111`, we see it's the maximum value that can fit:

```
    MSB                                 LSB
    31                                    0
    RRRR RTTT LLLL LLLL LLLL LLLL LLLL LLLL
    0000 0000 1111 1111 1111 1111 1111 1111
              ^
            Ok
```

> also (in this case) the autobahn implementation

I did not check the Autobahn implementation. I stumbled upon this while writing CppWAMP unit tests that used the maximum possible rawsocket length. I'm surprised nobody else discovered this before, and I suspect there may be bugs for this corner case in several implementations.

oberstet:
actually, you are right, I didn't pay close enough attention! the spec says:

> The three remaining octets constitute an unsigned 24 bit integer that provides the length of transport message payload following, excluding the 4 octets that constitute the prefix.

this only work up to 8M - not for 16M _maximum_, because an _actual_ length of 16M cannot be represented in 24 bits (in the standard unsigned int. repr.)

algo, the spec text might be more explicit .. as in saying **"provides the _actual_ length _(smaller than or equal to the announced maximum)_ of transport message payload"**

too bad, my fault :(

----

rgd fixing, there are a couple of ways .. with tradeoffs:

1. disallow `LENGTH == 15` altogether (but keep the 24 bits of actual length as is)
2. change the semantics/definition of `LENGTH`, decreasing it by `1`, either 2.a) selectively only for 16M (like your PR), or for 2.b) all `LENGTHs`
3. use one of the 5 reserved bits `R` in the message header `MSB LSB 31 0 RRRR RTTT LLLL LLLL LLLL LLLL LLLL LLLL`, thus increasing the bit length of the _actual_ message length by 1 to 25 bits (IOT: `MSB LSB 31 0 RRRR LTTT LLLL LLLL LLLL LLLL LLLL LLLL`)

Rgd 3., an implementation would only need to look at the new 25th `L` bit if `LENGTH == 15` _and_ all other `L` bits are `0` (to distinguish between a zero length payload, and a payload of 16M

I have some thoughts which one I'd prefer for some reasons to fix the issue, but maybe let's just first brainstorm and discuss ...

ecorm:
> 2. change the semantics/definition of LENGTH, decreasing it by 1 [snip] for 2.b) all LENGTHs

Your alternative 2b may be useful for possible C implementations that use fixed-size buffers that are powers of two in size, so that there's always a spare byte remaining in the buffer for null-terminating a string (in the case of JSON serialization).

oberstet:
> alternative 2b may be useful for possible C implementations ...

fwiw, I like that! 2b) had been on my fav list, but mainly because it is a _uniform_ solution that doesn't introduce a lot of special case treatment. I wasn't thinking of this C case, which is a very practical/real one IMO.

the downside of 2b) (as with others in my list) of course is: it is quite an invasive change. users relying on e.g. using `LENGTH == 1` for a usable maximum actual payload length of `2**10 octets` would have to change to `LENGTH == 2`  and such ...

let's wait for more comments if anyone feels to comment ... @KSDaemon @Jopie64  .. before "calling in a vote". obviously, this is a real bug that needs treatment in one way or other ...

ecorm:
> Rgd 3., an implementation would only need to look at the new 25th L bit if LENGTH == 15 and all other L bits are 0 (to distinguish between a zero length payload, and a payload of 16M

I like this option because it allows the full 16MiB frame length and does not break the current semantics of the LLLL bits in the handshake. Existing implementation that don't check your proposed 25th L bit (I would call it `LENGTH_CARRY`) would interpret a 16MiB frame length as zero length. Existing implementations need to fix this bug anyway, so they might as well start interpreting your proposed 25th L bit.

Existing applications that only use frame lengths less than 16MiB would not be affected by this new reserved bit being used.

oberstet:
Rereading the issue and our discussion from yesterday, I pretty much agree with your evaluation of option 3. in your last comments. For the reasons you describe, the option seems quite attractive to fix the bug.

The only thing that kinda slightly "unbeautiful" is that the L bits will not be consecutive anymore. I was expecting a need for new bits to the left of T bits, either to extend T or completely new bits. Well, I should get over such things;) "unbeautiful". Luckily I had reserved R bits in the first place, and using R bits is what they were reserved for, so if this is a good use (and it is IMO), then let's do it.

So I'd vote +1 for option 3.

Rgd the name of the 25th bit, I see why you came to "Carry" obviously, but it is again a slight asymmetry .. what if we ever needed a 26th L bit? other names that come to mind ... "25th L bit", "L-MSB", "L_24". but in any case, I don't care that much about this side aspect. the important part is: using the lowest index unused R bit as the new most significant 25th bit of L in a standard unsigned integer representation of the actual payload length.

konsultaner:
@ecorm @oberstet can you give me some time to think about that issue, because I have 3 implementations and I need to go through that issue.

konsultaner:
@ecorm great job finding that bug!

My solution is off topic, but I want to share it anyway. Maybe new thoughts might come up. I extended the handshake with an upgrade massage that extended the length from 2^24 - 1 to 2^32 -1. To allow larger payload sizes. This works pretty fine but has nothing to do with the bug.

Now maybe a 4th solution would be to instead of using the reserved bit, either just increase the length of the header in general or just add another bit in the end?

if this doesn't appeal to anyone i'm +1 for opinion 3.

KSDaemon:
Folks, give me a few days. I'll read all the findings today or tomorrow.

oberstet:
> Now maybe a 4th solution would be to instead of using the reserved bit, either just increase the length of the header in general or just add another bit in the end?
> if this doesn't appeal to anyone i'm +1 for opinion 3.

it does not appeal to me;)

actually, I would not want to extend the max beyond 16M .. even that is probably too large. if one want to transfer 16M, single messages are not the right way .. in my eyes .. but that would be up for a _different_ discussion

if you are also fine with opt. 3, the votes would currently stand at (if I'm correct, pls complain if not):

+3 (@ecorm @konsultaner @oberstet )

---

let's wait at least until @KSDaemon comes back ...

ecorm:
> Rgd the name of the 25th bit, I see why you came to "Carry" obviously, but it is again a slight asymmetry

I've thought about this, and "carry" implies some kind of arithmetic having been performed, which is not the case for us here. So I'm okay with `L_24` or whatever.

KSDaemon:
Folks I read all the discussions above and I think opt.3 is the best choice even if it is not as beautiful as we want. However, having backward compatibility is more important.

So as we all agree on the same, @ecorm can you update this PR with the required changes?

konsultaner:
> it does not appeal to me;)

@oberstet I know, we've alread had this discussion in another issue. I just couldn't find it anymore. I'm super happy with my upgrade header version. This is implementation specific and can be done by anyone else who needs it. :+1:

Jopie64:
For the record, and FWIW I don't really have an opinion for this since I never had to bother with raw sockets for WAMP :) Although option 3 looks OK for me too... And I agree with @oberstet that 16M is large for a single message, and not something what I'd expect for simple electronics that have to use raw sockets...

konsultaner:
> For the record, and FWIW I don't really have an opinion for this since I never had to bother with raw sockets for WAMP :) Although option 3 looks OK for me too... And I agree with @oberstet that 16M is large for a single message, and not something what I'd expect for simple electronics that have to use raw sockets...

@Jopie64 I mainly use WAMP to run SASS projects. In some ways how [meteor](https://www.meteor.com/) did it. So I need it to upload and download larger files.

Jopie64:
@konsultaner  Not sure what SASS means in this context, (other than a Stylesheet syntax), but you couldn't use progressive RPC for that?

KSDaemon:
I believe @konsultaner meant `SaaS` (Software as a Service)?

oberstet:
> but you couldn't use progressive RPC for that?

yeah, that would also be my view! in particular, since we now have progressive call results _and_ invocations. mass data transfer without proper integration into mass data storage (where the mass data usually comes from / goes to) doesn't cut it, and it is easy with progressive RPCs. need to "get_data()" return data from a 1GB file? np. etc etc ...

konsultaner:
> @konsultaner  Not sure what SASS means in this context, (other than a Stylesheet syntax), but you couldn't use progressive RPC for that?  

Sorry I mean Saas. 🙈

KSDaemon:
> yeah, that would also be my view! in particular, since we now have progressive call results and invocations. mass data transfer (network IO) without proper integration into mass data storage IO (where the mass data usually comes from / goes to) doesn't cut it, and it is easy with progressive RPCs. need to "get_data()" return data from a 1GB file? np. etc etc ...

Combining progressiveness with transparent payload... Mmm.... No need for additional encodes/decodes - that might be interesting combination for some cases :)

oberstet:
> No need for additional encodes/decodes - that might be interesting combination for some cases :)

yes, indeed!

> it could be made zero copy also. when using e.g. flatbuffers serializer which autobahn/crossbar has. but true zero copy down right into the NIC memory from user space requires more .. anyways;)

ecorm:
> So as we all agree on the same, @ecorm can you update this PR with the required changes?

Option 3 was @oberstet 's idea, so perhaps he would like to write up the PR himself? @oberstet , do you want to submit a new PR with your proposed fix? If you don't have the time, I could write something up and make you coauthor.

oberstet:
@ecorm you found the issue (which really is the tricky & crucial piece with such detail stuff) and also convinced me stupid that it is indeed an issue, plus submitted a PR in the first place, so I'd be happy if you adjust your PR to our conclusion, no need for me as co-author in git history or what. then yeah, actually I'm busy with some other stuff. thanks all for chiming in and contributing to a good result! besides the result, the working collaboration is great to see again=)

KSDaemon (changes requested):
Let's rewrite this with the agreed decision.

rfc/text/advanced/ap_38_transport_rawsocket.md
```diff
@@ -221,16 +222,32 @@ The serialized octets for a message to be sent are prefixed with exactly 4 octet
 {align="left"}
         MSB                                 LSB
         31                                    0
-        RRRR RTTT LLLL LLLL LLLL LLLL LLLL LLLL
+        RRRR XTTT LLLL LLLL LLLL LLLL LLLL LLLL
 
-The *first octet* has the following structure
+The *three least significant octets* constitute an unsigned 24 bit integer that provides the length of transport message payload following, excluding the 4 octets that constitute the prefix.
```
ecorm, line 227: I start with the description of the L bits here so that the description of the X bit below is easier to understand.

ecorm:
@oberstet , @konsultaner , @KSDaemon , @Jopie64

Changed text to reflect oberstet's option 3. See `Files Changed` tab.

oberstet (approved):
lgtm

KSDaemon (approved):
👍🏻 Great!

KSDaemon:
I'll merge now!

oberstet:
awesome! seems to be properly auto-deployed: https://wamp-proto.org/wamp_ap_latest_ietf.html#section-7.1-61

KSDaemon:
Auto-deploy rocks! :)

PR #511, opened 2023-11-22
Elaborate on subsequent progressive invocation options

### LINKED ISSUE #479: Propagation of frozen options in progressive call invocations
According to the recently-added section [Freezing of Options in Progressive Call Invocations](https://wamp-proto.org/wamp_latest_ietf.html#section-11.2-74), the Dealer must apply the options of the initial CALL request to all subsequent INVOCATION messages of the same request. It then gives `disclose_me=true` as an example where all subsequent INVOCATION messages must contain the caller identity.

It also states:

>  Not having to repeat (and not being able to change) options is more in tune with the concept of a media stream where options are set up initially, and the source (Caller) only needs to keep uploading more data thereafter.

So this makes things more efficient for the Caller, but not for the Dealer and Callee which must still communicate these repeated options.

**I'm therefore wondering if the all options (except for `progress`) should be omitted on the Dealer->Callee leg as well in subsequent INVOCATION messages for the sake of efficiency.** This would transfer the burden of remembering the original options to the Callee.

My client implementation (CppWAMP) keeps records of active progressive invocation calls in memory, and it already stores the options of the original CALL.

I believe the only other WAMP implementation currently supporting progressive invocations is the one by @KSDaemon (Wampy.js?)

ecorm:
Fixes #479

### DIFF
--- a/rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
+++ b/rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
@@ -448,3 +448,5 @@ For reference, here is a list of options that are frozen upon the initial progre
 - `CALL.Options.rkey`
 - `CALL.Options.runmode|string`
 - `CALL.Options.timeout|integer`
+
+These rules concerning progressive call invocation options also apply to custom options that are not part of this specification.
--- a/rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
+++ b/rfc/text/advanced/ap_10_rpc_progressive_call_invocations.md
@@ -428,7 +428,14 @@ For example, if `disclose_me=true` was specified in the initiating progressive c
 (within the same call) shall be considered by the *Dealer* to implictly have `disclose_me=true` as well. That is, all
 `INVOCATION` messages associated with the overall request shall contain caller identify information.
 
-Note that any option (besides `progress`) can be omitted altogether in subsequent progressive call invocations. Not having to repeat (and not being able to change) options is more in tune with the concept of a media stream where options are set up initially, and the source (*Caller*) only needs to keep uploading more data thereafter.
+Any option besides `progress` SHOULD be omitted altogether by the *Caller* in subsequent progressive call invocations. Not having to repeat (and not being able to change) options is more in tune with the concept of a media stream where options are set up initially, and the source (*Caller*) only needs to keep uploading more data thereafter.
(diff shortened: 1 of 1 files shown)


ecorm:
Here is a list of CALL options, for reference:

- `CALL.Options.disclose_me|bool`
- `CALL.Options.ppt_cipher|string`
- `CALL.Options.ppt_keyid|string`
- `CALL.Options.ppt_scheme|string`
- `CALL.Options.ppt_serializer|string`
- `CALL.Options.receive_progress|bool`
- `CALL.Options.rkey`
- `CALL.Options.runmode|string`
- `CALL.Options.timeout|integer`

oberstet (approved):
I'm +1 on this. It is actually a good restriction, and one that is lenient .. if that's the right word. because of "ignore", not "bail out".

> one minor note: so this "ignore" _any_ option but `progress` applies both to spec defined options _and_ custom options (`x_*`)? fine for me! could be made explicit in the text.

ecorm:
> one minor note: so this "ignore" any option but progress applies both to spec defined options and custom options (x_*)? fine for me! could be made explicit in the text.

Added a sentence to make it explicit for custom options.

oberstet:
so from my side, +1 RTFM .. maybe wait for at least one more vote from an non-author implementor? I'll add the flag.

> sidenote/OT: I'm also going to provide feedback and try a summary of a "step 1" rgd the other issue rgd limit URIs .. not now though ..

ecorm:
> sidenote/OT: I'm also going to provide feedback and try a summary of a "step 1" rgd the other issue rgd limit URIs .. not now though ..

I just wrote a big brain dump regarding the limit URIs issue. I suggest you read it first in case it overlaps with what you have in mind. Or it may trigger some different ideas on your part.

oberstet:
> > sidenote/OT: I'm also going to provide feedback and try a summary of a "step 1" rgd the other issue rgd limit URIs .. not now though ..
> 
> I just wrote a big brain dump regarding the limit URIs issue. I suggest you read it first in case it overlaps with what you have in mind. Or it may trigger some different/better ideas on your part.

thanks a lot! will do.

KSDaemon (approved):
👍🏻 LGTM!

KSDaemon:
I'm +1 for merging this! 
I'll try to update the Nexus while it is hot)

PR #513, opened 2024-01-24
correct company names - only formal corrections, no legal or substanc…

oberstet:
…e matter implications


oberstet:
I'm gonna merge this without review as it only and exactly correcting formal data rgd the entity responsible for publishing the spec and formally holding IP rights. no changes in substance matter or legal content whatsoever.

PR #516, opened 2024-03-25
Mis-spelling "canceLLing"

SunFulong:
This can be a bit terrible, because there are spelling errors in defined URI, please decide how to do this.

### DIFF
--- a/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
+++ b/rfc/text/advanced/ap_09_rpc_progressive_call_results.md
@@ -321,7 +321,7 @@ Such cancellation when the caller leaves addresses a potential security vulnerab
 
 **Callee Leaving**
 
-The *Dealer*'s behavior for when a *Callee* leaves or disconnects during a progressive results call shall be the same as in a basic, non-progressive call. That is, the *Dealer* sends an ERROR message to the *Caller* with the `wamp.error.canceled` URI. See [Callee Leaving During an RPC Invocation] (#rpc-callee-leaving) under the Basic Profile.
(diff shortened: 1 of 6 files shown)


oberstet:
there is no spelling error: https://www.merriam-webster.com/grammar/canceled-or-cancelled

SunFulong:
OK, I learned the American English.

But in [https://wamp-proto.org/wamp_latest_ietf.html#section-6.4](Section 6.4), here is a double-L "cancelled"

oberstet:
> But in [Section 6.4](https://wamp-proto.org/wamp_latest_ietf.html#section-6.4), here is a double-L "cancelled".

uups, ok! good catch!

I'd argue that consistency is more important, so A.E. flavor with 1 "l", so what you've found in your last post would be an error. but I'm not a native speaker, so. @ecorm what's your opinion?

oberstet:
I should have read more careful, I definitely think this is an _error_:

> then the Dealer SHALL return an ERROR message back to the Caller with a `wamp.error.cancelled` URI. 

as the "ll" appears in the URI `wamp.error.cancelled` .. which should be consistently single "l". that would make it consistent.

changing A.E. to B.E.: I'm not a native speaker, so what is actually "better" or what is definitely beyond me. BUT: at this point, with multiple implementations factually using single "l", I would strongly say: let's stick with single "l" just to keep everything non-breaking is much more important!

SunFulong:
> keep everything non-breaking is much more important

Ok, I agree with this.

My personal opinion is: US English is better because much more IT word are defined with it, such as we using `color` instead of `colour`, but you're right, if a `bug` using widely, then it is a `feature`, like HTTP header `referer` not `referrer`.

ecorm:
I agree with not breaking existing implementations.

PR #517, opened 2024-03-25
Client may abort a WAMP session by sending an ABORT message.

SunFulong:
As [Section-4.1.3](https://wamp-proto.org/wamp_latest_ietf.html#section-4.1.3) said:
> Both the Router and the Client may abort a WAMP session by sending an ABORT message.


KSDaemon (approved):
👍🏻 Thnx for the fix!

PR #518, opened 2024-03-26
Adding terminology page into full spec.

SunFulong:
Why not include the terminology into the full specification?


oberstet:
great, you've found this terminology page=) IMO a quite important one. so thanks for fixing the link and adding it to ToC in more places!

https://github.com/wamp-proto/wamp-proto/blob/master/rfc/text/wamp-terminology.md

PR #519, opened 2024-03-26
Re-Structure preview

SunFulong:
Hello, this PR have no content modify, only re-structures.

IMO, some chapters in current basic profile should extract out to the protocol spec level, so I moved them out, and move old "Basic Profile" title before "Message Definitions".

This is just a preview to discuss, if accept, I will rename BP files from 01.


oberstet:
Thanks for thinking about overall "structure" of the spec text! Improving readability and comprehensibility for new implementors is quite important! And a fresh, new look as you can bring to the table plus actual concrete proposals for improvements (as in this PR) is highly welcome!

As far as I see, this PR is really "only" restructuring existing contents - neither removing or adding any contents. Highly welcome, I appreciate, and to me, it looks like an improvement (and my problem in judging is likely anyways that I cannot have an independent fresh look anymore .. obviously)

So let's see what others say;)

SunFulong:
@oberstet Unfortunately, I'm not a native English speaker and writing content in English looks very unprofessional. We have also discussed some details before, such as adding additional algorithms to the CRA chapter. I also tried to write it, but finally gave up.

Therefore, the work I can do is probably correct errors, make minor adjustments to the content and structure, and come up with some new ideas.

oberstet:
> I also tried to write it, but finally gave up.

don't worry, I am pretty sure we can collaborate, I could improve english text / language to some point (not a native speaker also), and there are a few native english speakers in the WAMP implementors community which might chime in as well .. just saying .. pls don't hesitate to come up with text if you feel sth is missing or what ..

SunFulong:
In my work, I have already used the WAMP protocol in production projects, across WEB(PWA), App(navite) and Backend, and the router and client were all implemented from scratch. I have used Python, JavaScript, and now I am using Rust language for implementation.

As a private request, if you could add my name to the list of contributors, it would make it easier for me to promote this protocol among my social circles.

oberstet:
from my perspective, sure! pls see https://github.com/wamp-proto/wamp-proto/issues/520

KSDaemon:
I had a quick look: well — why not. Maybe it will be clear. It is important not to loose any parts during movements. So @SunFulong If you'd like — please go ahead with restructuring and let's discuss the PR Ready version.

Jopie64:
@SunFulong 
> Unfortunately, I'm not a native English speaker and writing content in English looks very unprofessional.

Not sure whether it is accessible in China, but I seriously think ChatGPT could help here. 😊 Just try to ask it to translate your Chinese text into professional English RFC-ready text.

SunFulong:
> Not sure whether it is accessible in China

The answer is NO, the ChatGPT blocks access from China, but Google Translate works.

SunFulong:
> no contents is removed, no new contents added, right?

Almost, changed some accidentally markdown tag, e.g. \*Caller\*s to \*Callers\*.

In fact, I want to make structural like this:

- Profiles
  - BP vs AP
  - Basic Profile
    - Messages
    - Session
    - etc
  - Advanced Profile
    - Messages
    - Session
    - etc

But, the level is too deep, so let the `BP` and `AP` as top-level, messages which follows the `BP` or `AP` as sub-level, the rest as top-level. Then I moved `BP vs AP` to Introduce chapter.

Don't you think if using `BP vs AP`, `BP`, `AP` as chapter separator is something strange?

rfc/text/advanced/ap_31_authentication.md
```diff
@@ -1,4 +1,4 @@
-# Authentication Methods {#auth-methods}
+## Authentication {#authorization}
```
KSDaemon, line 1: I think this should use the same name and anchor for correctness :)
```suggestion
## Authentication {#authentication}
```

KSDaemon (commented):
I have a look at changes. While in general everything looks good, I'm only worrying about too nested levels (5). I think that is too deep and it might be hard to navigate and even display such trees in user-friendly and neat way. Maybe we can somehow avoid such a deep nesting and move (possible with renames) some chapters up?

oberstet:
> Don't you think if using BP vs AP, BP, AP as chapter separator is something strange?

The overall guiding idea (originally) was:

* The BP is supposed to encompass only the (but all of) essentials of WAMP in _one_ part, and be quick and easy to implement
* The AP is supposed to be "optional" _per-part_, and implementors should be able to choose & select the parts of the AP wanted. And each AP part should be orthogonal / self-contained in general, but there _can_ be cases where one AP part depends on other AP part to be implemented as well

Thus, if I (as an implementor) want to start quick & easy, I want to read only that spec ... and not be bothered with or have to read or filter to get all info I need (for the BP only)

-----

Having said that, _how_ to achieve above goal as editors of the spec is up to debate of course! ;)

I am definitely open to listen and discuss whatever editorial / structural changes in the spec (both AP and BP) serve that goal! And I don't have a fixed opinion .. and I think I couldn't have one, since I am much too deep into all of it, and thus lost the "fresh look" / eyes which a newcomer that only wants the quick / easy stuff (BP) would have anyways.

KSDaemon:
Absolutely agree with @oberstet regarding
> The overall guiding idea (originally)

It would be awesome if we could keep that idea and just make it cleaner and more understandable :)

rfc/text/advanced/ap_31_authentication.md
SunFulong: Discuss this later, let's discuss how the full structure first.

SunFulong:
> The overall guiding idea (originally)

I totally understand this, but imagine if instead of putting AP after BP, we broke it up and mixed it up:

1. Forget that we also have 2 independent profile specifications, we only have the *all-in-one* one;
2. Don’t say BP or AP, just add the “(optional)” prefix or suffix to the chapter or section title if it is AP;
3. Filter out the optional section to independent AP profile, and the rest is the BP profile.

KSDaemon:
An idea:
We can modify our building process and instead of building one spec generate 2 specs:
1. The whole one
2. Only BP Spec
3. I think only AP is useless without BP

Both specs will consist of the same building blocks-chapters.
I believe this is a matter of creating 2 different reference-holding TOC files + some minor changes in building scripts (I can help with that if needed)

SunFulong:
> I think only AP is useless without BP

BP is the master spec, AP can be extended opt-in spec, we can have more than one extended specs.

oberstet:
my view:

> we can have more than one extended specs.

we could, but why?

having all advanced parts (which are optional and orthogonal as far as possible) which are _"official"_ in _one_ spec is IMO simpler and more clear.

otherwise, implementors will / might come up and publish other APs

and while implementors can of course extend and modify all of WAMP - both BP and AP - as they like, such material MUST NOT be misunderstood / confused with "official" BP or AP

currently, we have 3 documents coming from BP and AP material:

grafik

thinking about it: basically the only change at that top level that I would support at this point of the discussion would likely be:

drop 1 document, the combined spec with both of BP and AP

IMO the simplest structure would be: exactly 2 documents,

1. one BP-only, and
2. one with AP without BP-pieces

thus, an implementor that wants to go the "minimal + quick + easy" tour would implement the BP spec 1. and would only need to read that one document, and an implementor who wants it all would pick-and-implement parts from the AP spec _in addition_

KSDaemon:
well, I personally most of the time use full spec — because it is convenient to jump through different sections in one doc rather than navigating between different documents. I understand that might be just a matter of personal preference. But even context search for a term in one doc is way simpler than between a few documents (for example when you need to search for an exact uri across the whole spec).

SunFulong:
> drop 1 document, the combined spec with both of BP and AP ("WAMP Specification (full)")

-1 if only one AP, but +1 if multiple extensions.

> we could, but why?

IMO, one BP and multiple extensions simpler and more clear.

Imagine, if we split AP to `AA`, `AdvPubSub`, `AdvRPC`, `MetaAPI` extensions, then, I made a implementation, I could name it is `"A WAMP router with AA, AdvRPC extensions"`, but if only one AP, how to name? `"A WAMP router with particular AP (Chapter 2, 3, 7)"`?

KSDaemon:
> "A WAMP router with particular AP (Chapter 2, 3, 7)"

Unfortunately, this is not solvable for 100% as we all are humans and not robots...
Why? Let's take an example of progressiveness: there are progressive calls and progressive call results. Both these features are logically close enough to be specified in one extension. But one could implement only progressive call results and not implement progressive calls. And that is ok. But this breaks your idea of `"A WAMP router with AA, AdvRPC extensions"`.
At the same time, I don't like the idea of having a separate extension for every feature.

Some of the implementations just mention/list the supported features in their docs. And I like that! Unfortunately, not all implementations have this in their readme, so you have to dig into the code — that's not dev-friendly.

SunFulong:
> But one could implement only progressive call results and not implement progressive calls.

Even though, implementer can also say "A WAMP router with AA (full), AdvRPC (part)", separated extension is a edge, person can pike the edge instead of full AP. I don't believe the AP can have a final version, it will grow and grow more new features, so no one can say "with full AP", but extension can be final.

oberstet:
> I could name it is "A WAMP router with AA, AdvRPC extensions", but if only one AP, how to name?

just list the names of the AP features supported using the feature names as specified in the AP spec:

https://wamp-proto.org/wamp_ap_latest_ietf.html#name-feature-announcement

you might use the names as listed in that table, or you might use the formal identifiers, e.g. `progressive_call_results`

> Some of the implementations just mention/list the supported features in their docs. And I like that!

I agree with that!

ideally, we would have a big table on https://wamp-proto.org/implementations.html with columns per AP feature and check marks indicating support in the respective implementation in that row

-----

> the AP can have a final version, it will grow and grow more new features, so no one can say "with full AP", but extension can be final, or even have a new extension for existing extension.

yes indeed, saying "with full AP" doesn't make sense.

solution: don't say it! ;)

seriously: I think an implementation should list feature support in the docs, I agree with @KSDaemon 

having the features in the opening handshake technically agreed in a mandatory and verified way was one main reason to introduce opening handshake feature announcement in the first place

https://wamp-proto.org/wamp_ap_latest_ietf.html#name-feature-announcement

_any_ documentation, missing or not, is just "poetic words" .. what matters are the features as set in opening handshake

having said all that: unless and until we have a standard test suite and test all implementations vs all in a matrix fashion, and verify features announced by a given implementation actually _are_ implemented, and do that in an automated CICD fashion, I would not bet my life on it;)

BUT: I would argue, WAMP has all the fundamentals required and necessary to allow that!

SunFulong:
@oberstet , thanks for your long comment, but let's go back to the subject, what should we do next?

@oberstet means: drop the full spec, leave on BP and AP two spec;
@KSDaemon means: keep full spec;
@SunFulong means: keep BP spec, split AP to multiple extensions, and mix all of them into a new full spec.

Did I missing or mis-understand somethings?

SunFulong:
Re-pushed a version, the BPvsAP, BP and AP are treated as "TOP-TOP" level.

@oberstet , is this the presentation you wanted?

KSDaemon:
I tried to build the TOC in my head in runtime but failed %))) Need to build it and see what it will look like in real. 

The thing that I don't like: is that namings are not ordered. After we agree on the general structure — I suggest renaming them to keep ordering as it helps and simplifies navigation for spec editors (like me).

oberstet:
yes, I like that! specifically, splitting out `rfc/text/base/base_01_bp_vs_ap.md`, and then having that linked to in `rfc/text/basic.md` _at the top_ (right in the beginning, or preface), and also moving the `Terminology` to in-line. thank you! I think this whole PR is on a good way, and I can imagine this really improves readability for newcomers and others as well ... highly welcome!

SunFulong:
> I suggest renaming them to keep ordering

@KSDaemon Instead all sequence number, I prefer a partition sequence, e.g. PubSub using 30x RPC using 40x

> I think this whole PR is on a good way

@oberstet What should I do next? I think this is a ready stage, except file should rename.

SunFulong:
> Having some partitioning schemes is not clear from the file names itself

How about sub-directories plus seq-numbers?

oberstet:
> we should have an ordered sequence as it was before or remove it totally.

I understand your thinking and the "annoyance" (now) as shuffling around contents necessarily break the numbering match to order in TOC - which is unavoidable sooner or later, and consequently, I am +1 on **remove it totally**.

that is, have to numbers in files, only "proper names", refer those in TOC and be happy;)

KSDaemon:
Yep! Let's remove the numbers totally! :)

oberstet:
is there a FS where I can define `filename(<file>) = function(contents(<file>))`?  ;)  just kidding ..

KSDaemon:
Снимок экрана 2024-04-03 в 21 25 24

Heh.... This is how ToC looks now.... Not so great :(

SunFulong:
> Yep! Let's remove the numbers totally! :)

However, don't you think if we open the folder and the files are sorted alphabetically, it's another "confusion"?

KSDaemon:
> However, don't you think if we open the folder and the files are sorted alphabetically, it's another "confusion"?

I don't think so because the way files are sorted in directory — is a matter of your file browser/navigator but having numbers in file names — is the authors will.

SunFulong:
Rename job is done!

SunFulong:
Can we merge this PR and move to next round?

oberstet:
I am not sure I understand/follow all of your last comments (both of you), but relooking at the spec as it is published on the web site, in the context of our discussion here, can we discuss concrete proposal rgd document and TOC in literal form as comment here?

Eg how about

```
**COMBINED ONE SPEC:**

1. Introduction

2. Basic vs Advanced Profile

3. Basic Profile

4. Advanced Profile

```

as an implementor, after reading intro and BP vs AP, I might _only_ read BP and stop before AP -- or I might read everything?

KSDaemon:
@oberstet @SunFulong If we introduce one more nested level — I think we'll get exactly that structure!

KSDaemon:
I look at current ToC and think of next structure:
```
**COMBINED ONE SPEC:**

1. Introduction
2. Terminology
2.1. Fundamentals 
3. Basic vs Advanced Profile
4. Basic Profile
5. Advanced Profile
```

As terminology is a core fundamental block.

What do you think?

oberstet:
> If we introduce one more nested level — I think we'll get exactly that structure!

to make it concrete with an example, given top level TOC

```
**COMBINED ONE SPEC:**

1. Introduction

2. Basic vs Advanced Profile

3. Basic Profile

4. Advanced Profile

```

the section currently numbered **1.3**

```
[1.3. ](https://wamp-proto.org/wamp_latest_ietf.html#section-1.3)[Protocol Overview](https://wamp-proto.org/wamp_latest_ietf.html#name-protocol-overview)
```

which is part of **3.** BP would then need to be **3.1**

?

oberstet:
> As terminology is a core fundamental block.

yes it is. however, then why not make it part of **1. Introduction**, like similar to

```
**COMBINED ONE SPEC:**

1. Introduction
1.1. Fundamentals 
1.2. Basic vs Advanced Profile
1.3. Terminology

2. Basic Profile
2.1. Protocol Overview

3. Advanced Profile
```

I would argue "Fundamentals" or similar named very first section should really describe WAMP in most simple, casual terms.

Then say why there are 2 major parts following in the text .. that is BP and AP

And finally reference/list terms to prepare or/and come back ..

KSDaemon:
@oberstet I like your last suggestion! Yeah, all general things should be moved above any profiles as they are crucial for any profile

oberstet:
alright. let's nail it now that we are shuffling stuff around anyways.

I guess the last piece I am stumbling across is only rgd the _name_ of the very first section in Intro .. the *Fundamentals*.

In my eyes, a catchy title for "WAMP at a glance" or "WAMP quick and essential" or similar would be perfect ..

SunFulong:
I should -1 here.

The full spec is not a final delivery, if we downgrade all remaining chapters, there will be two separate specs with only one first level title

oberstet:
actually, for the very first "at a glance" intro section, we might steal back from eg

https://wamp-proto.org/intro.html#wamp

I would argue: part of the essential info **why** does it exist, and **what** can you (an app developer) do with it?

------

> The full spec is not a final delivery

yes it is: https://wamp-proto.org/wamp_latest_ietf.html

Jopie64:
> this should be a sketchy outline of the reason/features .. why does it exist anyways ... and similar

Simply "What is WAMP?" ?
or
"WAMP in a nutshell" ?

oberstet:
> "WAMP in a nutshell" ?

love it! =)

KSDaemon:
++ for the "WAMP in a nutshell" :)

oberstet:
would that work?

* **Part 1, 2 and 3**: https://wamp-proto.org/wamp_latest_ietf.html
* **Part 1 and 2**: https://wamp-proto.org/wamp_bp_latest_ietf.html
* **Part 1 and 3**: https://wamp-proto.org/wamp_ap_latest_ietf.html

SunFulong:
> The full spec is not a final delivery

I mean, not a final submit to IETF, but separated BP and AP is if we still plan to submit.

KSDaemon:
@SunFulong if you're saying about the difference in headers levels — that can be fixed during the build process. 
I don't mind about having only one full spec as @oberstet suggests. But in that case, it will be hard to submit only basic profile to IETF as a separate document.

oberstet:
> but separated BP and AP is if we still plan to submit.

IETF .. OASIS .. it would be nice! the latter could be easier, not sure. but hey, lots of talking and admin stuff just to get a RFC number? wtf, I have code to write;)

SunFulong:
See current push, I inserted HARD PART TITLE in full spec only

KSDaemon:
Снимок экрана 2024-04-03 в 22 49 02

Let's fix the ToC levels now?

SunFulong:
> Let's fix the ToC levels now?

Current build scripts won't fine for both separate and combined specs.

Let's do a final confirm, do we really downgrade profile part content level?

SunFulong:
> Introduction → Introduction

Tried over and over, my best idea is change first `Introduction` to `About`, how do you think?

oberstet:
uuh, lots of thoughts, attempts .. bit of chaos;)

first, let me preface: don't worry! this is pretty normal and expected. agreeing on anything between people, diverse as this group, coming from different backgrounds, having different focus and interests, is never easy and requires effort.

naturally. the hassles are unavoidable. fwiw, what we experience here is nothing compared what there was in WebSocket IETF WG ;)

-----

I tried to take a step back and focus on _what_ we can agree on, and figure out a concrete proposal for _that_ only.

of course not coming up with a "full proposal" for "everything" is a risk (later on, adding the parts taken out of focus for now might have ramifications upon the already "done" initial part"). well, that's life;) take risks.

anyways, what I _think_ we agree on (pls correct me / add as you see fit):

* we want one self-contained document for the BP contents only, and
* this document should only lay out the bare essentials ("minimum viable") of WAMP, and
* make it quick & easy for a potential implementer or newcomer to enter, and
* that one document should be formatted in a way so we have the option to submit it to IETF or OASIS

based on these goals, and also trying to take into account the other things we discussed (terminology, "in a nutshell"), and also some changes that appear to be desirable for me (from taking a step back and with a new look), here is my proposal for structure of above document:

grafik

KSDaemon:
Okay. I played a bit with it. And got to this point:

* All 3 docs have their abstracts (the difference is in adding profile names to BP/AP docs)
* BP and AP have their own introduction sections that are not copy/pasted to anywhere else.
* Only the full document has next sections:
  * WAMP in a Nutshell
  * Basic vs Advanced Profile
* BP and AP in the full spec are big top-level sections having all their details inside

The one question I still have in mind is the place for `WAMP in a Nutshell` section: I put it into full spec, but it might look better in BP/Introduction Section...

Here are screenshots of how it looks now.


Снимок экрана 2024-04-04 в 16 53 04

Снимок экрана 2024-04-04 в 16 58 15

Снимок экрана 2024-04-04 в 16 50 25

Снимок экрана 2024-04-04 в 16 57 41

Снимок экрана 2024-04-04 в 16 51 48

Снимок экрана 2024-04-04 в 16 57 56

oberstet:
> If this looks good — I can push this for review (probably in other PR)

mmmh, for my eyes, tbh, I don't think this is an improvement over the current text:(

E.g. "Abstract" for me is the same as "WAMP in a nutshell" and "BP vs AP" is an administrative frontmatter which should be outside the TOC.

In general, IMO we should only concentrate on BP document, because I think we even disagree on that, not even about the number of documents.

*Do we want 1 document, 2 documents or 3 documents?*

Once we agree, depending on the outcome, can we then _only_ discuss the _contents_ of the one _Basic Profile_ document and content ordering _only up_ "Publish & Subscribe" and "Remote Procedure Calls"?

What _content_ do we want a newcomer to read, in which order?

------

E.g. for me, it doesn't make sense to talk about "Protocol Overview" before "Building Blocks".

And in fact, before talking about these things, IMO the most important _guiding_ aspects should be introduced first:

* design paradigms
* security model

as well as a "common language" defined

* terminology

and only after that

* protocol building blocks
* protocol overview

described.

At this point, PubSub + RPC are the big _two_ chunks to be discussed which make up WAMP, and each should be introduced in _three_ sections corresponding to the respective three main actions in each of PubSub and RPC.

Anyways, just my opinion. However, as long as we don't agree I think it is a waste of time to work it into PRs.

meejah:
I would put most effort into the "Abstract" (this is where you get people excited enough to read more ... or lose them) and then "Introduction" (where you get a few more minutes / paragraphs to further hook anyone who didn't already leave).

After that, it's about well-organized and easy-to-find information for anyone who is already interested or "in" to WAMP etc.

A quick search didn't turn up where the words "1. Abstract" in the above screen-shots came from, but I can try editing etc if interested.

meejah:
Yes, I would certainly highlight the security aspects in "abstract", including "end-2-end message encryption". Despite renewed emphasis on "security" this is still a pretty unique feature for _any_ routed protocol, let along a real-time one like WAMP.

oberstet:
> highlight the security aspects in "abstract", including "end-2-end message encryption".

yes, agreed! unfort., only portions of it currently have text in AP spec

https://wamp-proto.org/wamp_ap_latest_ietf.html#name-advanced-security-features

I started and have half-done/ready text .. but ...

and the feature (e2e payload encr.) while having a working impl. in Crossbar.io/Autobahn lacks any other impls (as far as I know) currently. it has impls. in multiple AutobahnXXs ..

on the other hand CB/AB impls also have prototype impls. of the final stage of WAMP .. which isn't e2e in itself, but decentralized management of the trust _relations_ established between e2e encrypting WAMP clients/app endpoints (the apps that do PubSub and RPC and which want the payload in those events/calls to be protected even from the routing infra ..)

KSDaemon:
I think there is somewhere e2ee description that I was working on (IIRC in your fork @oberstet ). Maybe we can push it to this repo and continue working... 

I hope I'll get some stamina to implement e2ee in some of the wamp implementations under my wing...

oberstet:
> I think there is somewhere e2ee description that I was working on (IIRC in your fork @oberstet ). Maybe we can push it to this repo and continue working...
> 
> I hope I'll get some stamina to implement e2ee in some of the wamp implementations under my wing...

that would be fantastic! both of it! collaborating on

* polishing/nailing the E2E spec proposal text, _and_
* your implementation in one of your WAMP implementations and testing interoperability with AB/CB

and "testing interoperability" / "one of your implementations" .. any of or even all of the following would be perfect!

* your WAMP _client_ implementation, testing vs CB router implementation of E2E, and
* your _router_ WAMP implementation, testing vs AB client implementation of E2E, and _ultimately_,
* testing CB router implementation with your router implementation, for **E2E over rlinks** !

> fwiw, should you decide to do it, I would be highly interested in collaborating! obviously this would be an effort and PRs independent of this one.

oberstet:
one more comment, in general, "working on spec" _and_ "testing interop", with _different_ implementations and operators is in my eyes really the only way of hashing it out, nailing the spec, shaking out bugs, and all of it.

without _demonstrating_ different implementations and different operators working together, it's just words on paper ...

with that, words will be scrutinized and polished _by necessity_ and proven _by demonstration_!

SunFulong:
@oberstet Let's discuss this version:

Moved some sections from BP to BASE;
Downgrade 1 level for all profile sections.

And I tried move `BP vs AP` to frontmatter, but it seems current `xml2rfc` does not support.

SunFulong:
ping @oberstet @KSDaemon ...

oberstet:
I don't support changing and moving around sections unless/until we do have reached sufficient consensus

my own proposal is here: https://github.com/wamp-proto/wamp-proto/pull/519#issuecomment-2036777138

I lost sight of other's proposals, consolidated ones.

IOW: how do we discuss restructuring proposals so they are easy to track and compare?

right now, it's a big mess .. in my eyes ... at least I don't have a clue what is actually proposed (on the table) .. I don't have time to wade through pages and pages of changes in a PR which is sprinkled with various other changes as well

I would suggest to cut proposed changes into smaller and logically coherent PRs.

eg 1 PR with all structural changes, but only those, and only _after_ we agreed on the structure.

I don't know how to best discuss TOC changes (the structure&titles) .. some kind of table like .. but how to allow multiple people to submit their proposed changes, and still have only 1 big table with alternatives which are easy to compare ... mmmh .. any ideas?

SunFulong:
Sure, I don't mind reopen much PRs with small change, and we can using one TOC only issue or discuss to trace structure change

oberstet:
thank you!

>  reopen much PRs with small change

yes, smaller PRs and self-contained PRs, as in "one topic / thing" at a time would be highly welcome!

of course _creating_ those will be more work on your side.

BUT:

you want to get those changes _merged_.

and for that, we need consensus, and that is _much_ easier with smaller self-contained change packets (PRs)

changing the overall TOC / structure has yet an additional challenge: it is not just substance matter, but the ordering depends a lot on "perspective & taste", and everyone has a fav opinion ;)

plus the challenge of how to easily _discuss_ structural/TOC changes

PR #525, opened 2024-04-13
Fill full columns name, and remove unrelated column.

SunFulong:
My last PR today, it's time 2:50 a.m. in my timezone (UTC+08:00) now.

Leave one discuss here:

I want to resort AP's chapter orders in:

1. Session
2. PubSub
3. RPC

Both sections and this table, how do you think?


oberstet (approved):

KSDaemon (approved):
Kind of looks good! I'll merge

PR #526, opened 2024-04-21
Use correct error message format when dealer/broken dont support disclose_me

om26er:
That was likely overlooked when writing.


oberstet:
yeah, good catch! the `request_type` with the WAMP message type code for the original request, which in the 2 cases were 16 and 48

https://github.com/crossbario/autobahn-python/blob/f38f16ba28fa253dee951068cc729089c88d857d/autobahn/wamp/message.py#L1455C1-L1456C32

oberstet (approved):

KSDaemon (approved):
👍🏻

PR #529, opened 2024-06-22
docs: add octamp wamp

cydrickn:


oberstet:
great, thanks for contributing! I'm +1 on merging, seems legit including license, lets wait for at least one more upvote from another project member.

> my only comment: I think it would be nice and good for potential users of your project to read a short paragraph in https://github.com/octamp/wamp/blob/main/README.md of why octamp if "better" or different from other PHP router implementations ...

cydrickn:
> > my only comment: I think it would be nice and good for potential users of your project to read a short paragraph in https://github.com/octamp/wamp/blob/main/README.md of why octamp if "better" or different from other PHP router implementations ...

Thank, I already updated the README for short paragraph for comparison to other php framework

KSDaemon (approved):
Great! 👍🏻

KSDaemon:
@cydrickn I read the docs — it's a bit unclear, what slashes mean in the comparison table...
Снимок экрана 2024-06-23 в 11 08 15

KSDaemon:
But anyway — that is the question for the project itself :) and not for this PR. So merging!

oberstet:
> But anyway — that is the question for the project itself :)

funny, same here;) but it is fixed now it seems (using regular checkmark/cross now) ...

PR #531, opened 2024-06-24
ABORT a session if received YIELD msg w/ invalid INVOCATION.Request

SunFulong:
Need discuss.

Current spec not point what if dealer received a YIELD message, but its INVOCATION.Request is not issued by the dealer.

My point is to ABORT the session because it should be treat as MALFORMED message like non-sequential request ID.

### DIFF
--- a/rfc/text/basic/building_blocks.md
+++ b/rfc/text/basic/building_blocks.md
@@ -279,6 +279,7 @@ Following scenarios have to be considered protocol errors:
  - Receiving `REGISTERED` message, before session was established.
  - Receiving `UNREGISTERED` message, before session was established.
  - Receiving `INVOCATION` message, before session was established.
+ - Receiving `YIELD` message with invalid INVOCATION.Request.
  - Receiving message with non-[sequential](#session_scope_id) [session scope](#session_scope_ids) request ID, such as `SUBSCRIBE`, `UNSUBSCRIBE`, `PUBLISH`, `REGISTER`, `UNREGISTER`, and `CALL`. Note that there are exeptions for `CALL` when the _Progressive Call Invocations_ advanced feature is enabled. See the _Progressive Call Invocations_ section in the advanced profile for details.
  - Receiving protocol incompatible message, such as empty array, invalid WAMP message type id, etc.
  - Catching error during message encoding/decoding.


oberstet (approved):
> INVOCATION.Request is not issued by the dealer.

this is a misbehaving CALLEE, and I agree with your conclusion - in general, any peer misbehaving at this level should be treated rigorously / defensively, which means "kill the session" altogether .. there is no point in trying to work around .. it would expose us to even more issues likely ..

KSDaemon (approved):
Seems to be valid...

Jopie64:
Note that this could trigger a race. If the caller cancelled the the request, the router passes on the cancel to the callee and invalidates the invocation id.

If the callee sends the YIELD just before it received the cancel, the race will be triggered. 
This happened to me in the past with the crossbar router, which is a hard to find problem.

So if the router implements this strict behavior, it must take that race into account. I think this corner case should be mentioned in the doc and references to from this rule.

KSDaemon:
@Jopie64 Good point!

Jopie64:
> In this case - when the router passes a cancel to the callee - it can invalidate the invocation id noting it as being a canceled invocation's id.

That would indeed fix the race. But it would also introduce another problem: 
You can't store those id's forever because that would introduce a memory leak. So how would you purge that list? Would you introduce a timer? How long? If it's too short, network latency or processing time could trigger the race. If it's too long, the list could be exploited by bad actors to send a lot of calls, or maybe even naturally by rapid short running process invocations. 

Maybe something like:
- Router sends a ping and remembers when the ping is sent
- When corresponding pong is received, it can purge the invocationId's from before the corresponding ping is sent.

This would at least fix network latency races I think. (Although one would need to match a pong with a ping, and I think that's not possible atm...) 

That said, it also introduces extra logic and complexity and hence extra potentially exploitable risks. I'm not sure that would outweigh the risk of simply being tolerant of YIELD's with unknown invocation ID's.

oberstet:
I like the idea wrt ping/pong being used also to essentially establish "race epochs" .. that is, actively informing the other side about the ID up to which it has received and processed incoming messages ..

> When corresponding pong is received, it can purge the invocationId's from before the corresponding ping is sent.

What if we'd sent maximum _received_ IDs in the ping or pong payload along?

The other end could then clear and free up to the acknowledged ID - that is purge older IDs safely ...

oberstet:
> I'm not sure that would outweigh the risk of simply being tolerant of YIELD's with unknown invocation ID's.

on the one hand, I agree - it adds complexity. however, the added complexity is not for adding new functionality, but only to verify / double check .. "am I on the same page as my peer?" ..

Jopie64:
I don't think it's even necessary to add whole tables to the ping payloads. In my view, the only payload needed is an identifier to which the router can correlate the pong with the ping. Once the router received the pong, it knows for sure that the node is aware of every message received before the ping. So when the router sends a cancel for an invocation, and afterward a ping, the router can still expect YIELD's for that invocation, until it received the corresponding pong. Because then it knows it received the cancel as well.

But still I'm not convinced about the benefit on top of just ignoring YIELD's of expired invocations.
* If you look at misbehaving clients, they can misbehave in a lot of other protocol-valid ways. Sending YIELD with bogus ID's doesn't sound like a plausible way to mess up. But I can imagine it can get stuck in a loop sending valid CALL's.
* If you look at DOS attacks, then likewise you can DOS more easily in protocol-valid ways. Like setting up loads of CALL's. Or if you want to use YIELD, then just CALL yourself with a progressive RPC and DOS with YIELD's.
* Suppose this PING-PONG behaviour is introduced, then you might even have an extra attack vector. Just start and cancel a lot of CALL's, and postpone the PONG as long as possible, so that the router's list of expired invocations grows and grows.

I'm sorry to be pedantic. But I think it's good to think about implications of being strict vs tolerant before we force it upon lib builders...

SunFulong:
> If the caller cancelled the the request, the router passes on the cancel to the callee and invalidates the invocation id.
>
> If the callee sends the YIELD just before it received the cancel, the race will be triggered.

If caller cancelled, and:
1. callee support call canceling, the callee should reply either a `yield` or an `error`, when dealer got this message then invalidate the invocation id.
1. callee doesn't support call canceling, then should follow call timeouts rule, either a `yield` or an `wamp.timeout.error`.
1. callee doesn't support both of above, then this is a global problem: what if a callee never answer an invocation even if caller doesn't cancel a call?

Jopie64:
> callee support call canceling, the callee should reply either a yield or an error, when dealer got this message then invalidate the invocation id.

This is only the case when the cancellation mode of the `INTERRUPT` message is `kill`.
When the cancellation mode is `killnowait`, the router should not expect a response. (Even when it still can get one due to the race.)

Note that `killnowait` is the default when the invocation is cancelled due to a caller leave. See [docs](https://wamp-proto.org/wamp_latest_ietf.html#name-caller-leaving-during-an-rp).

> callee doesn't support both of above, then this is a global problem: what if a callee never answer an invocation even if caller doesn't cancel a call?

Good point. See also the last sentence of that paragraph:

> Whether or not call canceling is supported, the Dealer shall be prepared to discard a YIELD or ERROR response associated with that defunct call request.

I think this only strengthens my point.
For how long should it ignore YIELD's? Why add logic and complexity to remember defunct INVOCATION's for an unspecified time, instead of ignoring YIELD without valid ID altogether? What's the benefit?

SunFulong:
> For how long should it ignore YIELD's?

oberstet:
Let me add my current thinking .. fwiw, I think this discussion goes way beyond that simple one example, but touches on a general / fundamental issue .. which is very much worth discussing exhaustively - in the hope of resolving it once and for all;=

> But still I'm not convinced about the benefit on top of just ignoring YIELD's of expired invocations.

The benefit is "correct behavior", even in the presence of race conditions, versus "guess, ignore & hope for the best behavior".

In an **ideal world** (infinite speed & capacity), there are no races, and the correct behavior in my eyes is: *when a peer receives an invalid ID (such as in the example above for session-scope), immediately kill the session without sending even a single byte.*

In the **real world** (finite speed & capacity), and if the system is spatially distributed (hence there are non-zero link latencies), there are races necessarily, and the correct behavior in my eyes is: "when a peer receives an ID it cannot interpret and which _might_ be invalid, it must distinguish between *actually invalid* and *a consequence of a race*.

To do that, a peer must know the *moment of knowledge / valid_at / Kenntniszeitpunkt dt.* the other peer was at when it created and sent the message it sent. And that *valid_at* must be in reference to all data / timeframe it has processed incoming messages itself.

Thinking about it, I guess the most systematic and easy approach would be: each peer is adding a time reference (or outgoing message number or similar) to each message it sends e.g. in the WAMP message details.

The peer processing those messages can then use those to determine up to which point it has indeed knowledge of the state of the session of the other side, and finally add that again to message details of any message it produces and sends in response.

The receiving end can then correctly identify if:

a) there was a race, and hence a message with an ID like above should be ignored, OR
b) there was no race, and hence such a message is fundamentally bogus, and consequently the peer is broken, and the session must be killed

-----

I think above approach would allow systematic and correct handling of real world distributed messaging even in the presence of races.

> being strict vs tolerant 

I would call the 2 policies:

a) "guess, ignore & hope for the best behavior"
b) "strictly correct behavior"

Obviously, a) seems easier in the short term, and might be preferable to implementors, and we might allow it therefor.

b) seems correct and doable, and that is what I would personally aim for with my router.

Whether a router that implements "strictly correct behavior" does or does not accept any clients not implementing the same level is another question.

Again, personally, I would only care about b) clients and b) router, and with strict enforcement.

I am not interested in a) at all ...

> if the invocation ID is random I can treat the behavior as noise, but with sequential IDs it is a risk.

no, this is of course not true

PR #540, opened 2024-07-22
Add `wamplv` to list of client implementations

samangh:
This pull requests adds [wamplv](https://github.com/samangh/wamplv) to the list of WAMP client applications. This is a WAMP client written in LabVIEW, a data-flow driven language mostly used in the experimental sciences, engineering, robotics and real-time industrial control systems.

It would be great if we could include this in the implementation list, as far as I'm aware there are no other LabVIEW implementations of WAMP.

### DIFF
--- a/docs/implementations.rst
+++ b/docs/implementations.rst
@@ -61,6 +61,7 @@ wampcc_             C++                                 A C++ WAMP library that 
 WAMP_POCO_          C++                                 A fork of AutobahnCpp that is using `POCO <http://pocoproject.org/>`_ instead of Boost.
 wamped_             C++                                 Intended for running on `ARM mbed OS <https://www.mbed.com/en/>`_. Experimental.
 WAMPexClient_       Elixir                              Elixir client library that implements the complete client Basic Profile and much of the Advanced Profile.
+wamplv_             LabVIEW / G                         A WAMP client written in National Instrument's `LabVIEW <https://en.wikipedia.org/wiki/LabVIEW>`_, also available as a `VIPM package <https://www.vipm.io/package/samangh_lib_wamplv_wamp_client/>`_.
 WampSyncClient_     PHP                                 Synchronous (blocking) PHP client implementing Caller and Publisher roles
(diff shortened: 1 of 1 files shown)


KSDaemon:
Wow! That's awesome! I'm wondering, why you haven't come with this PR earlier!

KSDaemon (approved):
👍🏻

oberstet:
this is pretty cool! and I hope there might be some people get interested in the field of "IoT" and electrical engineering where "Matlab" is a big thing=)

Skimming over the repo, just 2 sidenotes / comments:

* `tcps://loclhost:8080/` as a URL .. the `s` for when running some raw TCP over TLS .. totally makes sense - but will confuse people;) as it is `tls` (or `ssl`) when refering to the raw byte stream without a specific application protocol
* the design of RPCs for integration into Matlab: is the design described somewhere? I only found https://github.com/samangh/wamplv/blob/master/docs/guide.md#asynchronous-calls. Eg it says: "The result of the call will be made available via a user event.". Why? A result and an event are vastly different things .. "normally" .. in WAMP in general. Anyways, I'm just a curious person, and the different ways/flavors of implementing WAMP in vastly different languages/runtimes is a particular interest;)

samangh:
@KSDaemon Thanks for merging this so quickly.

@oberstet Please note that is done LabVIEW, which is quite different to Matlab. They tend to be used in the same work environment (e.g. a physics or electrical laboratory), but are different languages. I've been thinking about making a Matlab client at some point, but I haven’t had the time so far.

With regards to your question, LabVIEW only has a few ways of having asynchronous callbacks and [user events](https://www.ni.com/docs/en-US/bundle/labview/page/creating-custom-events.html) are one of the more commonly used ones. An alternative is [notifiers](https://knowledge.ni.com/KnowledgeArticleDetails?id=kA00Z000000kFGfSAM&l=en-GB), which my library also fully supports. This way the user has full flexibility in how they integrate the library with the rest of their program.

Sadly there isn't much in the way of textual documentation, that's partially my fault and also due to the graphical nature of LabVIEW.

PR #545, opened 2024-10-31
Drop cancel(l)ed and cancel(l)ing's second L letter.

SunFulong:
After we discussed last time about `cancelled` should be `canceled`, I found another definition `cancelling` should be `canceling` which in autobahn implementation.

So, this PR, I replaced all of them in all texts, except word `cancellation`.

### DIFF
--- a/rfc/text/advanced/rpc_progressive_call_invocations.md
+++ b/rfc/text/advanced/rpc_progressive_call_invocations.md
@@ -20,16 +20,16 @@ and *Dealers* (`role := "dealer"`) via
         HELLO.Details.roles.<role>.features.progressive_call_invocations|bool := true
 
 Progressive call invocations can work only if all three peers support and announce this feature.
-In addition, *Callees* MUST announce support of the _Call Cancelling_ feature via
+In addition, *Callees* MUST announce support of the _Call Canceling_ feature via
 
 {align="left"}
-        HELLO.Details.roles.callee.features.call_cancelling|bool := true
+        HELLO.Details.roles.callee.features.call_canceling|bool := true
 
 
-As a consequence, *Dealers* MUST also announce support of the _Call Cancelling_ feature via
+As a consequence, *Dealers* MUST also announce support of the _Call Canceling_ feature via
 
 {align="left"}
-        WELCOME.Details.roles.dealer.features.call_cancelling|bool := true
+        WELCOME.Details.roles.dealer.features.call_canceling|bool := true
 
 The following cases, where a *Caller* sends a `CALL` message with `progress := true`, MUST be treated as *protocol errors*
 with the underlying WAMP sessions being aborted:
@@ -38,7 +38,7 @@ with the underlying WAMP sessions being aborted:
 - The *Dealer* did not announce the progressive call invocations feature during the `HELLO` handshake.
 
 Otherwise, in cases where the *Caller* sends a `CALL` message with `progress := true` but the *Callee* does not support
-progressive call invocations or call cancelling, the call MUST be treated as an *application error* with the *Dealer* responding
+progressive call invocations or call canceling, the call MUST be treated as an *application error* with the *Dealer* responding
 to the *Caller* with the `wamp.error.feature_not_supported` error message.
 
 **Message Flow**
--- a/rfc/text/basic/remote_procedure_call.md
+++ b/rfc/text/basic/remote_procedure_call.md
@@ -465,7 +465,7 @@ If either the *Dealer* or the *Callee* does not support the *Call Canceling* fea
 
 ## Callee Leaving During an RPC Invocation {#rpc-callee-leaving}
 
-After sending an INVOCATION message, if a *Dealer* detects that the *Callee* has left/disconnected without sending a final YIELD or ERROR response, then the *Dealer* SHALL return an ERROR message back to the Caller with a `wamp.error.cancelled` URI. The *Dealer* MAY provide additional information via the ERROR payload arguments to clarify that the cancellation is due to the *Callee* leaving before the call could be completed.
+After sending an INVOCATION message, if a *Dealer* detects that the *Callee* has left/disconnected without sending a final YIELD or ERROR response, then the *Dealer* SHALL return an ERROR message back to the Caller with a `wamp.error.canceled` URI. The *Dealer* MAY provide additional information via the ERROR payload arguments to clarify that the cancellation is due to the *Callee* leaving before the call could be completed.
 
 {align="left"}
         ,------.          ,------.          ,------.


oberstet (approved):

oberstet:
ah, another instance of *British English (BE)* vs *American English (AE)* difference.

> After we discussed last time  ..

yes, agreed, I believe it might be worth having this explicitly stated somewhere (in the spec?):

"we use *American English (AE)*"

----

> except word cancellation

fwiw, I think 

cancellation (BE) vs cancelation (AE)

is another instance.

if this is true, then using AE _everywhere_ (well, in all spec texts at least) would be desirable ..

oberstet:
I have no idea why/where/since-when this comes from

> sphinx.errors.SphinxWarning: Calling get_html_theme_path is deprecated. If you are calling it to define html_theme_path, you are safe to remove that code.

https://github.com/wamp-proto/wamp-proto/actions/runs/11614870567/job/32344052414?pr=545#step:15:3033

SunFulong:
> using AE everywhere

If so, should I change `canceling` (BE) back to `cancelling` (AE) in all text (not message defination)?

oberstet:
> If so, should I change ...

my opinion is: yes

I've added a "Needs consensus" label, to try to attract eyes, to call for consensus for:

**All spec text should use American English (AE) spelling.**

KSDaemon:
I'm definitely not a language expert) But... Of course, I think that spec should use one flavor everywhere.
Somehow, I personally prefer `cancelling` version :)

SunFulong:
Please do review again.

One more thing: how and when to change message's announces and error-uri to AE?

oberstet (changes requested):
sorry, I am confused .. _if_ we want AE everywhere, and _if_ I am right that AE is using 1 "l", as in

* canceling
* cancelation
* canceling
* canceled

then this last change doesn't match to that spelling ... ??

oberstet:
@SunFulong I'm sorry, I know this back-and-forth sucks!! in my experience with specs, this kinda "editor work" is unavoidable, we just need to fix it once and (ideally) make sure it is automated / not lurking in again ..

SunFulong:
@oberstet No need say sorry, *sublime text* makes it not too much cost, please review current version.

oberstet (approved):
looks good to me!

> `wamp.error.canceled` : I am +1 on this as well - but of course this might lead to regressions/issues when changing. e.g. in autobahn/crossbar.io, one way to handle it could be: always use the 1 "l" / new URIs when throwing errors, but accept both 2 and 1 "l" when receiving/consuming (map 2 "l" to 1 "l") or similar ..

Jopie64:
> If so, should I change `canceling` (BE) back to `cancelling` (AE) in all text (not message defination)?

It's the other way around. Single L is AE.

Looks like "cancellation" is an exception which is used with double L most of the time even in the US. (In c# there's a `CancellationToken` and a `Canceled` exception if I remember correctly.)

https://www.grammarly.com/blog/commonly-confused-words/canceled-vs-cancelled/

Also guess we should take extra care of `wamp.error.canceled` / `cancelled` since it's used in the protocol. I remember a discussion about that...

oberstet:
thanks @Jopie64 , I wasn't aware this rabbit hole is going even deeper ("Looks like "cancellation" is an exception"). 

if we could add some CI step to check spelling, and do that using some tool which is aware of all these shenanigans and which can do AE, that would save us time and conclude this (and similar) discussions ...

meejah:
I'm Canadian so maybe a bad sample (since we mix in a lot of "British" spellings as it is :wink: ) but both ways "look"/feel right to me.

Consistency is good, but for the price of breaking every implementation I'd probably choose to stick with the existing spelling.

oberstet:
> Consistency is good, but for the price of breaking every implementation I'd probably choose to stick with the existing spelling.

I would agree with that, definitely, and if so, that means we could look at this as a general policy which limits, but does so deliberately and specifically (should it be even mentioned somewhere in the spec? as in, "AE on new identifiers and all text", but "no fix of existing identifiers - only text"?).

rfc/text/advanced/rpc_progressive_call_invocations.md
```diff
@@ -20,16 +20,16 @@ and *Dealers* (`role := "dealer"`) via
         HELLO.Details.roles.<role>.features.progressive_call_invocations|bool := true
 
 Progressive call invocations can work only if all three peers support and announce this feature.
-In addition, *Callees* MUST announce support of the _Call Cancelling_ feature via
+In addition, *Callees* MUST announce support of the _Call Canceling_ feature via
 
 {align="left"}
-        HELLO.Details.roles.callee.features.call_cancelling|bool := true
+        HELLO.Details.roles.callee.features.call_canceling|bool := true
```
oberstet, line 26: please do not change this *attribute / identifier* - 
*Consistency is good, but backward compatibility too*

rfc/text/advanced/rpc_progressive_call_invocations.md
```diff
@@ -20,16 +20,16 @@ and *Dealers* (`role := "dealer"`) via
...
 
 
-As a consequence, *Dealers* MUST also announce support of the _Call Cancelling_ feature via
+As a consequence, *Dealers* MUST also announce support of the _Call Canceling_ feature via
 
 {align="left"}
-        WELCOME.Details.roles.dealer.features.call_cancelling|bool := true
+        WELCOME.Details.roles.dealer.features.call_canceling|bool := true
```
oberstet, line 32: please do not change this *attribute / identifier* - 
*Consistency is good, but backward compatibility too*

rfc/text/basic/remote_procedure_call.md
```diff
@@ -465,7 +465,7 @@ If either the *Dealer* or the *Callee* does not support the *Call Canceling* fea
 
 ## Callee Leaving During an RPC Invocation {#rpc-callee-leaving}
 
-After sending an INVOCATION message, if a *Dealer* detects that the *Callee* has left/disconnected without sending a final YIELD or ERROR response, then the *Dealer* SHALL return an ERROR message back to the Caller with a `wamp.error.cancelled` URI. The *Dealer* MAY provide additional information via the ERROR payload arguments to clarify that the cancellation is due to the *Callee* leaving before the call could be completed.
+After sending an INVOCATION message, if a *Dealer* detects that the *Callee* has left/disconnected without sending a final YIELD or ERROR response, then the *Dealer* SHALL return an ERROR message back to the Caller with a `wamp.error.canceled` URI. The *Dealer* MAY provide additional information via the ERROR payload arguments to clarify that the cancelation is due to the *Callee* leaving before the call could be completed.
```
oberstet, line 468: rgd the first change within this:

please do not change this *attribute / identifier* - 
*Consistency is good, but backward compatibility too*

oberstet (changes requested):
alright, I have reviewed the changes in light of @meejah 's comment and suggestion (which I support, not sure about "consensus" though ..):

*Consistency is good, but backward compatibility too*

consequently, all the changes to AE are good and desired (consistency), but any changes to existing *identifiers* are not good (backward compatibility beats consistency)

rfc/text/basic/remote_procedure_call.md
SunFulong: No, in this part, original is typo, your authbahn implementation is single 'l'.

rfc/text/advanced/rpc_progressive_call_invocations.md
SunFulong: No, in this part, original is typo, your authbahn implementation is single 'l'.

KSDaemon:
Folks, I also prefer to avoid breaking changes. 
And I checked wampy and nexus — these implementations use single `l` in `canceling` constants (in features and errors)

Jopie64:
Just my 2ct
I think it's important that docs mention explicitly where the specs deviate from AE because of backward compatibility. 

And also agree what @KSDaemon and @SunFulong say: when most implementations already use AE where the spec doesn't, I think it's better to change the spec _because_ of backward compatibility.
Maybe a little warning there that some implementations might use the BE spelling...

KSDaemon:
Well, we definitely can not have both spellings in the constants/code. So let's choose the most used one. That's why I went and checked the code. 
Nothing breaks right now, whatever the variant we choose. (broken can not be broken yet again, right?) But it will be good to state all these things: spelling, text, const, and caveats in the docs for the future. 
It's great that @SunFulong came and raised this question so we can fix it! :)

oberstet:
> But it will be good to state all these things: spelling, text, const, and caveats in the docs for the future.

absolutely. if that had been written down, it might have saved us all a bit time, and it pays off quickly, so I am definitely ++1

